PR #11, opened 2025-01-24
Feat/clin 3377 interpretation germline

creativeyann17:
- [x] Add Postgres DB  + repo
- [x] extract user info from token
- [x] Disable Starrocks if not used
- [x] escape HTML
- [x] generate JS client + OpenAPI DOC

**Tests**
- [x] Backend starts in local
- [x] Postgres starts and tables exist
- [x] Get a keycloak token
- [x] Requests to API work
- [x] history table works as expected 
- [x] Write unit test
- [x] Write integration test

**Endpoints**
- [x] GET interpretation germline
- [x] POST interpretation germline
- [x] GET interpretation somatic
- [x] POST interpretation somatic
- [x] GET pubmed

Problems:
- ~manual uuid cause gorm has issue retreiving it from postgres~ auto UUID working
- ~manual date cause same reason as above~ auto dates working
- using `,` string as array cause gorm cant transform them
- mappers model <=> DOA are boilerplate but work for complex mappings like pubmed
- we need a better way to handle errors, bad request, not found using gin context + custom middleware


backend/internal/client/pubmed_client.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil, err
+	}
+	defer res.Body.Close()
+	data, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != http.StatusOK {
+		log.Print("Error while fetching citation from pubmed: ", res.StatusCode, string(data))
+		return nil, nil
```
jecos, line 52: Is it ok to return no errors?

backend/internal/repository/interpretations.go
```diff
@@ -0,0 +1,195 @@
...
+
+func (r *InterpretationsRepository) FirstSomatic(sequencingId string, locus string, transcriptId string) (*types.InterpretationSomatic, error) {
+	var dao types.InterpretationSomaticDAO
+	if result := r.db.
+		Table(types.InterpretationSomaticTable.Name).
+		Where("sequencing_id = ? AND locus_id = ? AND transcript_id = ?", sequencingId, locus, transcriptId).
+		First(&dao); result.Error != nil {
+		err := result.Error
+		if !errors.Is(err, gorm.ErrRecordNotFound) {
+			return nil, err
```
jecos, line 166: Add some context to the error => fmt.Errorf

backend/internal/repository/interpretations.go
```diff
@@ -0,0 +1,195 @@
...
+		dao.CreatedByName = existing.CreatedByName
+	}
+ 	var res types.InterpretationSomaticDAO
+	result := r.db.Table(types.InterpretationSomaticTable.Name).
+		Where("sequencing_id = ? AND locus_id = ? AND transcript_id = ?", interpretation.SequencingId, interpretation.LocusId, interpretation.TranscriptId).
+		Assign(dao).
+		FirstOrCreate(&res)
+	err = result.Error
+	if err != nil {
+		return err
```
jecos, line 191: Add some context to the error => fmt.Errorf

backend/internal/server/middelwares.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/golang-jwt/jwt"
+)
+
+type DecodedJWT struct {
+	Name string
+    jwt.StandardClaims
+}
+
+func ExtractJWTInformation() gin.HandlerFunc {
```
jecos, line 17: I think you dont need to decode the token by yourself. gin-keycloak can get information tou need here :
[See this](https://github.com/tbaehler/gin-keycloak?tab=readme-ov-file#how-to-get-the-keycloak-claims-eg-sub-mail-name)

jecos (changes requested):

backend/internal/server/middelwares.go
jecos: If it works, dont forget to remove unused deps

backend/internal/client/pubmed_client.go
creativeyann17: I had some un-usual response while trying unknown citation ID, i'm gonna improve the error handling

backend/internal/server/middelwares.go
creativeyann17: Oh nice ... so far not working ... maybe a bug in the lib code
https://github.com/tbaehler/gin-keycloak/blob/3f68b881eb27a74eee8667cd60ebcffea7cf55e4/pkg/ginkeycloak/gin_keycloak.go#L264

creativeyann17: ok it's working

celinepelletier (approved):

backend/internal/client/pubmed_client.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil, nil // equivalent to NOT FOUND during tests
+	}
+	client.cacheMutex.Lock()
+	client.cache[id] = citation
+	defer client.cacheMutex.Unlock()
+	return citation, nil
+}
+
+func (client * PubmedClient) getCitationByIdFromCache(id string) (*types.PubmedCitation, bool) {
+	client.cacheMutex.RLock()
```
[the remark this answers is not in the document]
creativeyann17: - the read lock on the map line 72 works in tandem with the write from line 67
- go will handle several requests we dont want them to access that map at the same moment
- the `defer` is used to assure the mutex will unlock whatever the function return, panic etc ... 

I could actualy switch lines 67 and 66, defer should be called asap.

Note: that cache is dumb ... will increase in memory and never be cleared up by a timeout or anything, also not shared between the pods, this is a cheap way to avoid spamming pubmed API.

jecos (approved):

PR #114, opened 2025-03-27
feat: SJRA-105 add statistics endpoint for occurrences

celinepelletier:


backend/internal/repository/starrocks_repository.go
```diff
@@ -298,6 +300,45 @@ func (r *StarrocksRepository) AggregateOccurrences(seqId int, userQuery types.Ag
...
+			selectedPanelsTables := getDistinctTablesFromFields(userQuery.GetFieldsFromTables(types.GenePanelsTables...))
+			for _, panelsTable := range selectedPanelsTables {
+				tx = tx.
+					Joins(fmt.Sprintf("LEFT JOIN %s %s ON %s.symbol=cf.symbol", panelsTable.Name, panelsTable.Alias, panelsTable.Alias))
+			}
+		}
+		addWhere(userQuery, tx)
+
+	}
+	return tx, part, nil
```
jecos, line 323: We are nos supposed to have the same query that aggregate except n groupby and min, max instead of count()?

backend/internal/types/table_field.go
```diff
@@ -11,7 +11,8 @@ type Table struct {
 }
 
 const (
-	ArrayType = "array"
+	ArrayType   = "array"
+	NumericType = "numeric"
```
jecos, line 15: for my concenr, what is exactly the purpose of this field?

jecos (changes requested):

backend/internal/types/table_field.go
jecos: 👍

backend/internal/repository/starrocks_repository.go
celinepelletier: duplicated function removed

jecos (approved):

PR #149, opened 2025-04-11
✨ Add new colors

oliviercperrier:


francisl (commented):
If we go that route that mean:
1. when we overwrite the colors per themes, we will load both colors

Wouldn't be better to have a template and load it directly the common from the portals theme?

oliviercperrier:
> If we go that route that mean:
> 
> 1. when we overwrite the colors per themes, we will load both colors
> 
> 
> 
> Wouldn't be better to have a template and load it directly the common from the portals theme?

@francisl These common colors will always be the same for all portals. Thats why i put them in a common css file.

francisl (approved):
🍻

PR #250, opened 2025-05-15
🐛 Fix interpretation tooltip glitch

oliviercperrier:
The tooltip was wrapping the entire Dialog component, causing the tooltip to open unexpectedly.


francisl (approved):
:shipit:

PR #286, opened 2025-05-27
feat: SJRA-267 add migration script to create clinical tables in PostgreSQL

celinepelletier:


backend/scripts/init-sql/migrations/000004_create_clinical_tables.up.sql
```diff
@@ -0,0 +1,489 @@
+-- Value Sets ---------------------------------------------------------------
+
+CREATE TABLE IF NOT EXISTS "sexe"
```
kritchie, line 3: I think it should be `sex` instead.

kritchie (approved):

PR #320, opened 2025-06-09
:sparkles: feat: SJRA-274: Add `task_id` into Occurrences

kritchie:
https://d3b.atlassian.net/browse/SJRA-274

## Context

With https://d3b.atlassian.net/browse/SJRA-265  completed, the clinical data model was made available locally (with synthetic data) for development. 

We need to update the current Occurrence definition to be scope on tasks as well as sequencing experiment. 

## Contribution

This PR brings the following changes:
- Adds the `task_id` field to the Occurrence data model
- Updates the test fixtures and associated models


backend/internal/types/occurrence.go
```diff
@@ -95,6 +96,12 @@ var SeqIdField = Field{
 	CanBeFiltered: true,
 	Table:         OccurrenceTable,
 }
+var TaskIdField = Field{
+	Name:          "task_id",
+	CanBeSelected: false,
```
celinepelletier, line 101: should be true since it's in the default fields

kritchie: Addressed in https://github.com/radiant-network/radiant-portal/pull/320/commits/b9bc588237a596ff60f1b5d35f38868730b75f8e

PR #337, opened 2025-06-19
feat(case): SJRA-152 completed case exploration view list

lflangis:
image


frontend/apps/case-exploration/src/feature/cells/priority-cell.tsx
```diff
@@ -0,0 +1,32 @@
...
+  code: 'asap' | 'routine' | 'stat' | 'urgent';
+};
+
+const colors: Record<string, IndicatorProps['variant']> = {
+  asap: 'amber',
+  routine: 'slate',
+  stat: 'red',
+  urgent: 'blue',
+};
+
```
francisl, line 15: if we hardcode colors like this, it would not be possible in the future to adapt to new portals easily, it should be themable.

frontend/apps/case-exploration/src/feature/cells/status-cells.tsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const colors: Record<string, BadgeProps['variant']> = {
+  draft: 'slate',
+  'on-hold': 'yellow',
+  active: 'blue',
+  revoke: 'red',
+  completed: 'green',
+  incomplete: 'orange',
+};
```
francisl, line 16: same thing here

frontend/components/base/data-table/cells/date-cell.tsx
```diff
@@ -0,0 +1,11 @@
+import { formatDate } from 'date-fns';
+
+type DateCellProps = {
+  date: string;
+};
+
+function DateCell({ date }: DateCellProps) {
+  return <div className="text-muted-foreground">{formatDate(date, 'yyyy-MM-dd')}</div>;
```
francisl, line 8: wouldn't here be better to localize the date?

francisl (changes requested):

PR #366, opened 2025-07-04
✨ Update DataTable header

oliviercperrier:
DO NOT MERGE YET, need to talk to lucas about header bg color which currently has an opacity and that doesnt work with pinned row (we see the column in the background when hovered)

https://github.com/user-attachments/assets/94a569f8-a10b-4d2d-8324-25d521030214


francisl (approved):
:shipit:

PR #367, opened 2025-07-04
feat: SJRA-294:  Changes for cases list

celinepelletier:
- Add primary condition id and name in case result
- Include count in the list (renamed search) response
- Remove count endpoint for case

⚠️ Breaking changes, need to update UI


lflangis (approved):

frontend/apps/case-exploration/src/components/cases/cases-tab.tsx
```diff
@@ -137,30 +119,32 @@ function CasesTab() {
     },
   );
 
+  console.log('data', data);
```
celinepelletier, line 122: remove console.log

PR #440, opened 2025-07-15
:sparkles: feat/SJRA-428 Add Clinvar RCV endpoint

kritchie:
## Context

In https://github.com/radiant-network/radiant-portal-pipeline/pull/90 we added the Clinvar RCV summary table. 

This table enables adding variants information using the ClinVar RCV database. 

We need to make this available in the API. 

## Contribution

A new endpoint was added: `/variants/germline/:locus_id/conditions/clinvar` to fetch the RCV rows associated with a variant. 

No pagination or sorting is implemented for this endpoint as the resulting data is generally a few lines long.


backend/internal/types/clinvar_rcv.go
```diff
@@ -0,0 +1,36 @@
+package types
+
+import (
+	"time"
+)
+
+type Submission struct {
```
celinepelletier, line 7: Where do you use Submission ?

PR #570, opened 2025-08-04
:sparkles: feat: SJRA-548 Add 'has_variants' flag in case search response

kritchie:
## Context

To be able to filter cases for which there are (or not) variants, we need to add that information to the case search endpoint. 

This PR adds the `has_variants` (bool) to the `api/cases/search` endpoint response.  

## Contribution

The `has_variants` is calculated using the `staging_sequencing_experiment` table, both the following conditions have to be true for a case to have variants:
- At least 1 row exist for a case in the `staging_sequencing_experiment`.
- The `ingested_at` column contains a valid date. 

New payload: 

```
{
    "case_id": 1,
    "mrn": "MRN-283775",
    "priority_code": "routine",
    "status_code": "active",
    "case_analysis_code": "WGA",
    "case_analysis_name": "Whole Genome Analysis",
    "case_type": "germline_family",
    "requested_by_code": "CHUSJ",
    "requested_by_name": "Centre hospitalier universitaire Sainte-Justine",
    "project_code": "N1",
    "created_on": "2021-09-12T13:08:00Z",
    "updated_on": "2021-09-12T13:08:00Z",
    "has_variants": true 
}
```


frontend/api/configuration.ts
```diff
@@ -90,11 +90,10 @@ export class Configuration {
         this.basePath = param.basePath;
         this.serverIndex = param.serverIndex;
         this.baseOptions = {
```
lflangis, line 92: Si je comprendds bien, vue que headers est juste un spread de baseOptions.headers, je pense que ça ne veut plus la peine de spread les deux

```suggestion
        this.baseOptions = param.baseOptions
```

PR #585, opened 2025-08-11
:sparkles: feat: SJRA-598 Add CNV Occurrence endpoint

kritchie:
## Context

Germline CNVs are a new data set of Occurrences we need to add the Radiant system. 

This means the API needs to support listing and counting this new data set. 

## Contribution

The following PR adds the following endpoints routes: 

- `occurrences/germline/cnv/:sed_id/list`
- `occurrences/germline/cnv/:sed_id/count`

These endpoints support the usual filtering and sorting methods using `sqon` definition.


backend/internal/repository/germline_cnv_occurrences.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	tx := r.db.Table(
+		fmt.Sprintf("%s %s", types.GermlineCNVOccurrenceTable.Name, types.GermlineCNVOccurrenceTable.Alias),
+	).Where("seq_id = ? and part=?", seqId, part)
+
+	return tx, nil
+}
+
+func (r *GermlineCNVOccurrencesRepository) getPart(seqId int) (int, error) {
```
celinepelletier, line 83: I don't see any difference with the SNV one, so we could re-use the same instead of duplicating code

backend/internal/repository/germline_cnv_occurrences.go
```diff
@@ -0,0 +1,91 @@
...
+	utils.AddWhere(userQuery, tx)
+
+	var count int64
+	if err = tx.Count(&count).Error; err != nil {
+		return 0, fmt.Errorf("error fetching occurrences: %w", err)
+	}
+	return count, nil
+}
+
+func (r *GermlineCNVOccurrencesRepository) prepareListOrCountQuery(seqId int, userQuery types.Query) (*gorm.DB, error) {
```
celinepelletier, line 70: Maybe do a function that has in parameter the table object, so we can use it for both snv and cnv

backend/internal/repository/germline_cnv_occurrences.go
```diff
@@ -0,0 +1,91 @@
...
+
+func NewGermlineCNVOccurrencesRepository(db *gorm.DB) *GermlineCNVOccurrencesRepository {
+	if db == nil {
+		log.Print("GermlineCNVOccurrencesRepository: db is nil")
+		return nil
+	}
+	return &GermlineCNVOccurrencesRepository{db: db}
+}
+
+func (r *GermlineCNVOccurrencesRepository) GetOccurrences(seqId int, userQuery types.ListQuery) ([]GermlineCNVOccurrence, error) {
```
celinepelletier, line 32: Since the logic is exactly the same as in SNV, maybe refactor it with a generic typing to avoid duplication

celinepelletier (changes requested):

PR #655, opened 2025-09-02
feat: SJRA-634 add autocomplete for document search

celinepelletier:
Autocomplete to search on document_id, run_name, sample_id, patient_id, case_id, sequencing_experiment_id and task_id

Query SQL:

```
(SELECT "document_id" as type, id as value from `radiant_jdbc`.`public`.`document` WHERE CAST(id AS TEXT) LIKE '1%')
UNION
(SELECT "run_name" as type, run_name as value from `radiant_jdbc`.`public`.`sequencing_experiment` WHERE run_name LIKE '1%')
UNION
(SELECT "sample_id" as type, id as value from `radiant_jdbc`.`public`.`sample` WHERE CAST(id AS TEXT) LIKE '1%')
UNION
(SELECT "patient_id" as type, id as value from `radiant_jdbc`.`public`.`patient` WHERE CAST(id AS TEXT) LIKE '1%')
UNION
(SELECT "case_id" as type, id as value from `radiant_jdbc`.`public`.`cases` WHERE CAST(id AS TEXT) LIKE '1%')
UNION
(SELECT "seq_id" as type, id as value from `radiant_jdbc`.`public`.`sequencing_experiment` WHERE CAST(id AS TEXT) LIKE '1%')
UNION
(SELECT "task_id" as type, id as value from `radiant_jdbc`.`public`.`task` WHERE CAST(id AS TEXT) LIKE '1%')
ORDER BY value asc, type asc;
```

Example of response: 

```
[
   {"type":"case_id", "value":"1"}, 
   {"type":"document_id", "value":"1"}, 
   {"type":"patient_id", "value":"1"}, 
   {"type":"sample_id", "value":"1"}, 
   {"type":"seq_id", "value":"1"}
]
```


backend/internal/repository/documents.go
```diff
@@ -51,6 +53,63 @@ func (r *DocumentsRepository) SearchDocuments(userQuery types.ListQuery) (*[]Doc
...
+	*/
+	var autocompleteResult []AutocompleteResult
+	searchInput := fmt.Sprintf("%s%%", prefix)
+
+	subQueryDocumentId := r.db.Table(fmt.Sprintf("%s %s", types.DocumentTable.Name, types.DocumentTable.Alias))
+	subQueryDocumentId = subQueryDocumentId.Select("\"document_id\" as type, id as value")
+	subQueryDocumentId = subQueryDocumentId.Where("CAST(id AS TEXT) LIKE ?", searchInput)
+
+	subQueryRunName := r.db.Table(fmt.Sprintf("%s %s", types.SequencingExperimentTable.Name, types.SequencingExperimentTable.Alias))
+	subQueryRunName = subQueryRunName.Select("\"run_name\" as type, run_name as value")
```
kritchie, line 81: Shouldn't this be `\"run_name\" as type, id as value` like the other sub queries ? 🤔

kritchie (approved):
Single question. 
Otherwise this looks good 👍 
🚀

backend/internal/repository/documents.go
celinepelletier: No because it's an autocomplete to find run names and ids so for example you type "123" and it will propose 
- patient_id 1234
- run_name 123AJDH
- case_id 123
etc...

PR #1244, opened 2026-04-28
feat: SJRA-1393 add transcript_id to default fields for snv occurrences list endpoint

LysianeBouchard:
# feat(SJRA-1393):  include transcript_id by default in SNV occurrences list endpoints

## 📌 Context

The somatic and germline SNV occurrence list endpoints should return transcript_id by default.

This field is required for creating an interpretation, so it should be readily available to the frontend—without needing to explicitly request it or call the expanded occurrence endpoints.

**Before this PR:**
The somatic list occurrences endpoint did not include transcript_id at all.
The germline list occurrences endpoint only included transcript_id if it was explicitly requested by the frontend.

**After this PR**:
Both endpoints always return transcript_id in the response, even if it is not explicitly requested.

## 📝 Changes

  - Added `transcript_id` to the default fields of `SomaticSNVOccurrence` and `GermlineSNVOccurrence` types (`backend/internal/types/`)
  - Updated integration tests and repository/handler tests to cover the new default field                                                                                                                                               
  - Regenerated Swagger documentation (`backend/docs/`) via `make doc`                                                                                                                                                                  
  - Regenerated the TypeScript client (`frontend/api/`) and Python client (`cli/python/`) via the `make generate-client-typescript` and `make generate-client-python` commands from the root Makefile                                   
                                                                                                                                                                                                                                        
  > Note: changes in `backend/docs/`, `cli/python/` and `frontend/api/` are **auto-generated** 

## ☑️ TO-DOs

<!-- Tasks that still need to be completed before merging. -->

## 🧪 Tests

  - Updated integration tests: `germline_occurrences_integration_test.go`, `somatic_snv_occurrences_integration_test.go`                                                                                                                
  - Updated repository tests: `germline_snv_occurrences_test.go`, `somatic_snv_occurrences_test.go`
  - Updated handler tests: `handlers_germline_snv_occurrences_test.go`, `handlers_somatic_snv_occurrences_test.go` 

In the integration tests, for both germline and somatic, there is one case with a non-empty transcript_id (included in the response) and one with an empty transcript_id (omitted from the response).

I did not add extra handler test cases specifically for default fields or missing values, as these edge cases don't seem to be covered. However, the mock repository was modified to returns a non-empty transcript_id to better reflect typical usage.

For the repository tests, I added an assertion for transcript_id where relevant.

## 🔗 Related Issues

<!-- Begin JIRA Issues -->
- [SJRA-1393](https://d3b.atlassian.net/browse/SJRA-1393)<!-- End JIRA Issues -->

## 👀 Notes for Reviewers

<!-- Anything reviewers should pay special attention to. Questions, caveats, tricky parts, etc. -->

-  Sometimes, there is no transcript_id. Assumed that it is ok to omit it from the response in this case.

- It seems like there is no somatic cases when testing locally (manually starting the application), so for somatic endpoing, we only cover through unit tests. 
I think it would take a considerable amount of time to add somatic data and that it would deserve it's own ticket and prioritization.  But if you think the testing effort is not good enough, let me know.

[SJRA-1393]: https://d3b.atlassian.net/browse/SJRA-1393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kritchie (approved):

LysianeBouchard:
Je suis arrivée à tester end-to-end le endpoint somatique via le sandbox (merci @kritchie  pour la suggestion).   Il y a une erreur quand je clique sur le case somatique, mais ne semble pas lié à mes changements.  Je vois ceci:  
```
React Router caught the following error during render TypeError: Cannot read properties of undefined (reading 'observed_phenotypes')
```

Cependant, si je clique sur bouton "voir variants" associé au case, je peux loader la page de variants correctement. Je vois que le output retourné par endpoint list occurrences est correct, i.e. on voit bien la liste de variant avec chacun leur champ transcript_id.

PR #1270, opened 2026-05-06
feat: sjra-1429 update `GetVariantClinvarConditions` to return conditions for a single `clinvar_id`

kritchie:
## 📌 Context

When return the Clinvar RCV conditions in `GetVariantClinvarConditions` based on a `locus_id`, it was possible to obtain more than one `clinvar_id`. This is problematic for the UI. 

The updated method JOINS the Clinvar RCV with the Variants table to get the appropriate `clinvar_id` for which to return the Clinvar RCV values.

## 📝 Changes

- Update the SQL query in `GetVariantClinvarConditions` to `JOIN` the Variant table with the Clinvar RCV table to return only the values for a single `clinvar_id`
- Adjust the return payload of the `/variants/germline/{locus_id}/conditions/clinvar` to set the `clinvar_id` at root level.
- Add a small utility to be able to prefix columns when writing SQL queries. (Useful when there's a bunch of columns in the Select and we still want to use the `Alias` of the table). 

## 🧪 Tests

- Update the related unit tests.
- Tested using the Sandbox. 

### Manual Python client smoke tests (not commited)

Using a local implementation (not commited) of tests using Python CLI to validate the changes on a running Sandbox with the updated API. This validates 2 things, the feature and that the clients (at least the Python client) were updated correctly. 

ℹ️ This is not added to this repository since are already versioning the integration tests and the Postman tests spec. 

## 🔗 Related Issues

- https://d3b.atlassian.net/browse/SJRA-1429

## 👀 Notes for Reviewers

The new payload for `/variants/germline/{locus_id}/conditions/clinvar` is:

```
{
   "clinvar_id": "abc123",
   "conditions": [{...}, {...}]
}
```


AltefrohneGaelle (approved):

backend/internal/types/clinvar_rcv.go
```diff
@@ -20,6 +20,11 @@ type ClinvarRCV struct {
 	Origins              JsonArray[string] `json:"origins,omitempty" gorm:"type:json"`
 }
 
+type ClinvarVariantConditions struct {
+	ClinvarId  string       `json:"clinvar_id,omitempty"`
+	Conditions []ClinvarRCV `json:"conditions"`
```
celinepelletier: JsonArray[ClinvarRCV] probably would be better

celinepelletier (commented):
Small comment but optional

backend/internal/types/clinvar_rcv.go
kritchie: Adressed in [34f4a59](https://github.com/radiant-network/radiant-portal/pull/1270/commits/34f4a5984be5e16bce6b290e7896b95b5ce01512)

celinepelletier (approved):

PR #1290, opened 2026-05-14
feat: SJRA-1490 add cases task with occurrences endpoint

celinepelletier:
# feat(api): SJRA-1490 add cases tasks-with-occurrences endpoint

## 📌 Context

The Variants tab for a case currently filters occurrences by `seq_id` only, so when a sequencing experiment is reused across cases, occurrences leak into cases that have no annotation task of their own. Fixing this end-to-end requires the API to return, per (case, sequencing), the list of tasks whose task type actually produces the occurrences a given Variants sub-tab displays. This PR is the first of three sub-tasks (API list endpoint → API occurrence filters → UI rework) and delivers the new tasks-list endpoint.

## 📝 Changes

- New endpoint **`GET /cases/{case_id}/{seq_id}/tasks_with_occurrences?data_type=germline_snv|germline_cnv|somatic_snv`** that returns tasks attached to the (case, sequencing) pair whose task type produces occurrences of the requested `data_type`, sorted by `created_on DESC`. The mapping from `data_type` to task type is resolved server-side (germline_snv → `radiant_germline_annotation`, germline_cnv → `alignment_germline_variant_calling`, somatic_snv → `radiant_somatic_annotation`).
- Repository filter is `seq_id = ? AND (case_id = ? OR case_id IS NULL) AND task_type_code = ?`. The OR-NULL branch is what allows per-seq alignment tasks to legitimately appear for every case that reuses the sequencing, while case-scoped annotation tasks remain isolated to their case.
- New types: `types.OccurrenceType` enum + `TaskTypeCode() (*string, error)`, `types.TaskOccurrenceType` response DTO (with `validate:"required"` on always-present fields).
- Worker task-type constants consolidated to `internal/types` so both `cmd/api` and `cmd/worker` reference the same source of truth. New `AlignmentSomaticVariantCallingTaskTypeCode` constant.
- Worker map adjustments:
  - `CaseRelatedTaskTypes`: `+radiant_somatic_annotation`, `-alignment_somatic_variant_calling`.
  - `RequiresInputDocumentsTaskTypes`: `+radiant_somatic_annotation`.
  - `SingleAliquotTaskTypes`: `+alignment_somatic_variant_calling`.
- Postgres seed fix: `task_context` rows for task 74 (`radiant_somatic_annotation`) corrected to case-scoped (`case_id = 71` instead of `NULL`).
- Generated artifacts refreshed: `backend/docs/swagger.{yaml,json,go}`, `frontend/api/api.ts`, `cli/python/`. Postman collection updated with a new request under the Cases folder.
- New `add-endpoint` skill at `backend/.claude/skills/add-endpoint/SKILL.md` and shared backend rules at `backend/.claude/rules/universal.md`. `backend/CLAUDE.md` now points to them instead of duplicating a brief workflow.

## ☑️ TO-DOs

- [x] Confirm worker map reclassification of `alignment_somatic_variant_calling` (case-related → single-aliquot) does not break any prod batch validation. Worker tests are green but the semantics may matter operationally.

## 🧪 Tests

26 new Go tests + 2 security-test entries, all green via `make test`:

- **Types** — 5 tests in [`backend/internal/types/occurrence_test.go`](backend/internal/types/occurrence_test.go) covering `OccurrenceType.TaskTypeCode()` for each valid value, unknown input, and empty input.
- **Repository** — 7 tests in [`backend/internal/repository/task_test.go`](backend/internal/repository/task_test.go) for `ListTasksByCaseSeqAndTaskType`: happy path per data type, empty result, sort order, exclusion of tasks attached to a different case, and the **shared-sequencing scenario** (two cases reusing the same seq must both see the case-agnostic task but only the right case sees the case-scoped task).
- **Handler** — 9 tests in [`backend/internal/server/handlers_cases_tasks_test.go`](backend/internal/server/handlers_cases_tasks_test.go) using `assert.JSONEq` against expected JSON strings.
- **Integration** — 5 tests in [`backend/cmd/api/cases_tasks_integration_test.go`](backend/cmd/api/cases_tasks_integration_test.go) using the modern `testutils.RunTest` + `Need{Postgres: ReadPostgres}` API.
- **Security** — new route added to `Test_SecureRoutes` (asserts 401 without auth) and `Test_OpenFGA_Authorization` (asserts 200 with a `data_manager` JWT) in [`backend/cmd/api/integration_test.go`](backend/cmd/api/integration_test.go).

To run: `cd backend && make test`.

## 🔗 Related Issues

- [SJRA-1490](https://d3b.atlassian.net/browse/SJRA-1490)

## 👀 Notes for Reviewers

- The query semantics (`case_id = ? OR case_id IS NULL`) is the load-bearing part of the fix. Tests `Test_ListTasksByCaseSeqAndTaskType_CaseAgnosticTaskReturnedForBothCasesSharingSequencing` and `Test_ListTasksByCaseSeqAndTaskType_ExcludesTaskAttachedToDifferentCase` pin both branches of the OR.
- The route lives under `/cases/{case_id}/{seq_id}/tasks_with_occurrences` rather than `/tasks`, because the endpoint intentionally returns only the tasks that produce occurrences of the requested type — not the full task list for the (case, seq) pair. 
- Worker map changes reclassify `alignment_somatic_variant_calling` from `CaseRelatedTaskTypes` to `SingleAliquotTaskTypes` and add `radiant_somatic_annotation` to the case-related/input-doc maps. 
- The seed change in `backend/test/data/clinical/05_task_context.sql` corrects somatic annotation rows to be case-scoped, matching real ETL output. Verified that the pre-existing `Test_RetrieveCaseTasks_DeduplicatePatients` (case 71) still passes.
- The new `backend/.claude/skills/add-endpoint/SKILL.md` and `backend/.claude/rules/universal.md` files codify the workflow used here. Not a runtime change — feel free to skim and push back on the rules wording.

[SJRA-1490]: https://d3b.atlassian.net/browse/SJRA-1490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/cmd/worker/case_validation.go
```diff
@@ -90,29 +90,28 @@ const (
...
 
 var CaseRelatedTaskTypes = map[string]struct{}{
 	"family_variant_calling":            {},
 	"tumor_only_variant_calling":        {},
-	"alignment_somatic_variant_calling": {},
-	RadiantGermlineAnnotationTask:       {},
-	ExomiserTaskTypeCode:                {},
+	types.RadiantGermlineAnnotationTask: {},
+	types.RadiantSomaticAnnotationTask:  {},
+	types.ExomiserTaskTypeCode:          {},
```
LysianeBouchard, line 99: Pourrait être bien d'indiquer pourquoi on ajoute la contrainte.

Pour les cnv, il me semble qu'on assumait le type alignment_germline_variant_calling.  On ne l'ajoute pas?

backend/internal/repository/task.go
```diff
@@ -114,3 +116,39 @@ func (r *TaskRepository) GetTaskHasDocumentByDocumentId(documentId int) ([]*Task
 	}
 	return thd, nil
 }
+
+func joinTaskContextWithTask(tx *gorm.DB) *gorm.DB {
+	return tx.Joins(fmt.Sprintf("LEFT JOIN %s %s ON %s.task_id = %s.id",
+		types.TaskTable.Name, types.TaskTable.Alias,
+		types.TaskContextTable.Alias, types.TaskTable.Alias))
+}
```
LysianeBouchard, line 124: Il me semble avoir vu une fonction avec le même nom dans internal/utils/repository.go ... 
On ne peut pas utiliser celle-là parce qu'elle est faite pour un query à travers starrocks et non pas postgres, c'est ça?

backend/internal/types/occurrence_test.go
```diff
@@ -0,0 +1,49 @@
...
+	assert.Error(t, err)
+	assert.Nil(t, code)
+	assert.Contains(t, err.Error(), "not_a_real_type")
+}
+
+func Test_OccurrenceType_TaskTypeCode_Empty(t *testing.T) {
+	code, err := OccurrenceType("").TaskTypeCode()
+
+	assert.Error(t, err)
+	assert.Nil(t, code)
```
LysianeBouchard, line 48: Pas d'assert sur le error dans ce cas-ci comme pour le unknown?

backend/cmd/worker/case_validation.go
celinepelletier: So ETL uses this task to load the CNV but this task (as far as I understand it) is not linked to a case, it just needs the output of the sequencer (no family context, no phenotypes etc...) 
So if you add the given sequencing to another case, you should be able to see the CNVs as well because they weren't produced in the context of a specific case

backend/internal/repository/task.go
celinepelletier: Yeah.. I will create a Jira ticket to see how to deal with this concern (maybe we should request postgres through Starrocks only if we need to join with Starrocks tables, use direct requests on PostgreSQL for the rest)
I'm not sure how to organize repositories to make that easy to understand yet ...

backend/internal/types/occurrence_test.go
celinepelletier: I can add it but we already checked the error in the other test case

backend/internal/repository/task.go
LysianeBouchard: C'était vraiment juste une question pour ma compréhension personnelle, mais c'est une bonne idée pour le JIRA :) . Demande un peu de réflexion bien sûr.

backend/internal/types/occurrence_test.go
LysianeBouchard: Ok

backend/cmd/worker/case_validation.go
LysianeBouchard: Ok, de ce que je comprends, c'est assumé quand les CNV sont envoyés que tous les case qui pointent sur le sequencing associé doivent pouvoir les voir.  C'est ok dans ce cas j'imagine. Et j'imagine que c'est ok pour les crams dans les IGV tracks aussi.

LysianeBouchard (approved):

PR #1301, opened 2026-05-21
fix: SJRA-1494 Fix documents leak between cases sharing sequencings

celinepelletier:
# fix(case-entity): SJRA-1494 Fix documents leak between cases sharing sequencings

## 📌 Context

The case documents query (POST /cases/{case_id}/documents/search) fans out task_context rows across every case that owns the sequencing, because the shared join helper JoinTaskContextWithCaseHasSeqExp matches on sequencing_experiment_id only. When two cases share a sequencing, case-scoped task documents leak from one case to the other.

## 📝 Changes

- In [JoinTaskContextWithCaseHasSeqExp](backend/internal/utils/repositories.go#L179), add a `case_id` predicate to the `LEFT JOIN ON` clause so that `task_context` rows are paired with the matching `case_has_sequencing_experiment` row for the same case (or remain unmatched when `tctx.case_id IS NULL`).
- In [retrieveCaseTasks](backend/internal/repository/cases.go#L391), simplify the `WHERE` clause to only `chseq.case_id = ?` now that the per-case scoping happens in the join.

## ☑️ TO-DOs

- [ ] Code review

## 🧪 Tests

- Manually verify on a case that shares a sequencing experiment with another case: the document list should only include documents belonging to the requested case.
- Verify that tasks with no `task_context` (i.e. `tctx.case_id IS NULL`) are still returned (and their documents as well)

## 🔗 Related Issues

- [SJRA-1494](https://d3b.atlassian.net/browse/SJRA-1494)

[SJRA-1494]: https://d3b.atlassian.net/browse/SJRA-1494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/internal/repository/cases.go
+++ b/backend/internal/repository/cases.go
@@ -388,7 +388,7 @@ func (r *CasesRepository) retrieveCaseTasks(caseId int) (*[]CaseTask, error) {
 	tx = utils.JoinTaskContextWithSeqExp(tx)
 	tx = utils.JoinSeqExpWithSample(tx)
 	tx = utils.JoinSampleAndCaseHasSeqExpWithFamily(tx)
-	tx = tx.Where("(tctx.case_id = ? OR tctx.case_id IS NULL) AND chseq.case_id = ?", caseId, caseId)
+	tx = tx.Where("chseq.case_id = ?", caseId)
 	tx = tx.Select("task.id, task.task_type_code as type_code, task.created_on, task_type.name_en as type_name, group_concat(f.relationship_to_proband_code) as patients_unparsed, count(distinct spl.patient_id) as patient_count")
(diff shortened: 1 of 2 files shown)


backend/internal/repository/cases.go
LysianeBouchard: Hors scope de ce PR, mais je le note au passage : le Joins sur txStg n'est pas réassigné à la variable, alors que c'est fait ailleurs dans la fonction.

Peut-être un ticket à créer pour investiguer ça à part?

LysianeBouchard (approved):

PR #1303, opened 2026-05-21
fix: SJRA-1171 distinguish zero from null nb_snv on germline cnv occurrences

LysianeBouchard:
#  fix(germline-cnv):  always serialize nb_snv and improve nullability of other numeric fields (SJRA-1171)

## 📌 Context

A dash (`-`) was displayed in the `nb_snv` column of the UI in two distinct cases: when no data was available, and when no SNV was overlapping the CNV.  This was flagged as a bug — the ask was to display `0` when no SNV overlaps, and keep the `-` only for situations where we don't know whether overlapping SNVs exist or not.
  
After investigation, we realized the problem came from two layers: the`GermlineCNVOccurrence` API response and the ETL. Neither allowed distinguishing the two scenarios. The ETL is using a NULL value for both cases and the API is systematically omitting both 0 and NULL values.

After discussion, we decided to always display `0` in the UI — both when the nb_snv value is unknown and when there is no overlap. Distinguishing the two scenarios was considered but judged too costly on the ETL side (see this PR for more context: https://github.com/radiant-network/radiant-portal-pipeline/pull/169).

The ETL behavior is left as-is — it continues to write `NULL` in both cases — and the radiant-portal backend is updated in this PR to convert these `NULL` values to `0` on the wire so the frontend always receives an integer.

While at it, we also tightened how other optional nullable numeric fields are serialized so a real `0` is never confused with an unknown value (`NULL`).

  ## 🔄 Strategy change vs. initial PR
  The initial version of this PR used `*int` / `*float32` pointers for `nb_snv` so that `0` and `NULL` could be distinguished on the wire. That approach was dropped  for `nb_snv` = 0 only — see context above.  Pointer treatment is **kept** for the other nullable numeric fields, since for those a real `0` is meaningful.
  
  
### Backend type — types.GermlineCNVOccurrence
  - `NbSNV`: stays `int` but **`omitempty` removed** → always serialized; DB `NULL`
    is read as `0`, which is the desired UI behavior.
  - Converted to pointers (so real `0` is preserved and `NULL` is
    omitted): `Quality`, `BC`, `CN`, `SM`, `SVLen`, `RefLen`, `NbGenes`,
    `GnomadSC`, `GnomadSN`, `GnomadSF`.

  ### Test schema alignment
  - `germline__cnv__occurrence` table schema in `backend/scripts/init-sql/init_starrocks.sql`
    and `backend/test/data/sql/germline__cnv__occurrence.sql` aligned with the
    schema in radiant-portal-pipeline (adds `phased`, `cytoband`, `symbol`,
    `nb_genes`, `nb_snv`, `gnomad_*`, etc.).
  - TSV files `test/data/{simple,multiple}/germline__cnv__occurrence.tsv` extended
    with an `nb_snv` column containing both `NULL` and integer rows so both
    branches are exercised
   - also add and populate the `nb_snv` column in script `backend/scripts/init-sql/init_starrocks.sql` used for the docker compose test setup
  - `test/testutils/setup_starrocks.go`: support the literal `"NULL"` token in
    TSV data so it is inserted as a real SQL `NULL`.
    
  ### Misc
  - `Test_PostSequencingExperimentBatch_OK` is occasionally failing due to
  cross-package parallelism in Go. Best effort to reduce the probability of failure for this test.
  - `backend/README.md`: fix Keycloak role to assign manually (`api` → `radiant`).
    
## 🧪 Tests

 - **Repository** (`Test_GermlineCNV_GetOccurrences_NbSnv`): asserts the three
    branches — DB `NULL` → `0`, DB `0` → `0`, DB `1` → `1`.
  - **Handler** (`Test_CNVOccurrencesListHandler`): mocks two CNVs (one with
    `NbSNV=1`, one with `NbSNV=0`) and asserts `"nb_snv": 0` and `"nb_snv": 1`
    both appear in the JSON output — i.e. the field is never omitted.
  - **Integration** (`Test_CNVOccurrence_List`, `Test_CNVOccurrence_List_Filter_On_Chromosome`):
    expected JSON updated to include `nb_snv`.
  - Verified locally with docker-compose: rows where `nb_snv` is `NULL`  in
    the DB is displayed as `0` in the UI; non-zero values display correctly.
  - Double check that make doc + generator commands do not introduce changes
 
## 🔗 Related Issues

- [SJRA-1171](https://d3b.atlassian.net/browse/SJRA-1171)

## 👀 Notes for Reviewers

- For the flaky test Test_PostSequencingExperimentBatch_OK, we do a best effort to stabilize. It might not completely get rid of the problem, but we hope that it will reduce the probability of a concurrent test causing interference.   Apparently, go can run tests in parallel even though parallel mode is not specified if they belong to different packages. See this blog: https://bryce.is/writing/code/go-test-and-parallelism.html

[SJRA-1171]: https://d3b.atlassian.net/browse/SJRA-1171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/cmd/api/batches_integration_test.go
```diff
@@ -15,21 +15,22 @@ import (
 	"github.com/radiant-network/radiant-api/test/testutils"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"gorm.io/gorm"
 )
 
 func Test_PostSequencingExperimentBatch_OK(t *testing.T) {
-	testutils.ParallelTestWithPostgres(t, func(t *testing.T, postgres *gorm.DB) {
+	testutils.RunTest(t, testutils.Need{Postgres: testutils.ExclusivePostgres}, func(t *testing.T, env *testutils.Env) {
```
LysianeBouchard, line 21: We use sequential mode (ExclusivePostgres) because we write and claim a batch. Both operations can interefere with other tests.

backend/cmd/api/batches_integration_test.go
```diff
@@ -15,21 +15,22 @@ import (
...
+		repo := repository.NewBatchRepository(db)
 
 		router := gin.Default()
 		router.POST(
 			"/sequencing/batches",
 			server.PostSequencingExperimentBatchHandler(repo, &auth),
 		)
 
-		runDate := time.Date(2025, 12, 15, 12, 0, 0, 0, time.UTC)
+		runDate := time.Date(2024, 12, 15, 12, 0, 0, 0, time.UTC)
```
LysianeBouchard, line 33: Using a date < 2025-01-01  so the row falls inside auto cleanUp's preservation range.    

Apparently, tests from different packages can run concurrently (a Go test limitation) even though we are not using the parallel mode.  This will make the test a bit more robust to this, although it is not a perfect fix. 

We manually remove the batch after the test is done below, since it won't be covered by the auto cleanUp.

backend/scripts/init-sql/init_starrocks.sql
```diff
@@ -346,32 +346,46 @@ PARTITION BY (`part`)
...
-    `svtype`       VARCHAR(32),
-    `svlen`        INT,
-    `reflen`       INT,
-    `ciend`        ARRAY<INT>,
-    `cipos`        ARRAY<INT>
-)
-ENGINE=OLAP
-DUPLICATE KEY(`part`, `seq_id`, `id`);
+     part int(11) NOT NULL,
+     seq_id int(11) NULL,
```
celinepelletier, line 352: should not be nullable

backend/scripts/init-sql/init_starrocks.sql
```diff
@@ -346,32 +346,46 @@ PARTITION BY (`part`)
...
+     task_id int(11) NOT NULL,
+     cnv_id bigint(20) NOT NULL,
+     aliquot varchar(50) NULL,
```
celinepelletier, line 355: Same

backend/scripts/init-sql/init_starrocks.sql
```diff
@@ -346,32 +346,46 @@ PARTITION BY (`part`)
...
+     chromosome varchar(20) NULL,
```
celinepelletier, line 356: Same

backend/scripts/init-sql/init_starrocks.sql
```diff
@@ -346,32 +346,46 @@ PARTITION BY (`part`)
...
+     alternate varchar(20) NULL,
+     start int(11) NULL,
+     end int(11) NULL,
+     type varchar(10) NULL,
```
celinepelletier, line 360: Same, why did you put all the fields nullable ?

backend/test/data/sql/germline__cnv__occurrence.sql
```diff
@@ -1,35 +1,39 @@
...
-     `gnomad_sc`    INT NULL,
-     `gnomad_sn`    INT NULL,
-     `gnomad_sf`    FLOAT NULL
-)
-ENGINE=OLAP
-DUPLICATE KEY(`part`, `seq_id`);
\ No newline at end of file
+CREATE TABLE IF NOT EXISTS `germline__cnv__occurrence` (
+     part int(11) NOT NULL,
+     seq_id int(11) NULL,
```
celinepelletier, line 3: Same question here

backend/test/testutils/setup_starrocks.go
```diff
@@ -216,11 +216,11 @@ func createTableAndPopulateData(db *sql.DB, folderName string, file os.DirEntry)
 		// Convert to interface{} for stmt.Exec
 		args := make([]interface{}, len(values))
 		for i, v := range values {
-			//if strings.HasPrefix(v, "[") {
-			args[i] = fmt.Sprintf("array(%v)", v)
-			//} else {
-			args[i] = v // Database driver will handle type conversion
-			//}
+			if v == "NULL" {
```
celinepelletier, line 219: Starrocks wasn't able to handle NULL ?

celinepelletier (changes requested):

backend/scripts/init-sql/init_starrocks.sql
LysianeBouchard: Ok, I will change it.  

I used the same schema as in radiant-portal-pipeline, but finally, I think it might not be perfectly reliable ...

In this case, I think that I should probably use `NOT NULL`  for all columns matching a field marked as "required" in  the GermlineCNVOccurrence struct?

I will make sure do the same changes in radiant-portal-pipeline PR.

backend/scripts/init-sql/init_starrocks.sql
LysianeBouchard: See comment above. I used the schema in radiant-portal-pipeline thinking it was the source of truth. But apparently I was wrong. Sorry for that.

backend/test/testutils/setup_starrocks.go
LysianeBouchard: No :(

backend/scripts/init-sql/init_starrocks.sql
LysianeBouchard: Precision,  I changed the schema because I realize that it was not up to date. For exemple, the nb_snv column did not exist and the cnv_id column was called id ...

celinepelletier: It's a good call to take the one from the ETL since that's the source of truth, I'm wondering what so many fields are nullable tho.... seq_id being nullable doesn't make sense for me 🤔

backend/scripts/init-sql/init_starrocks.sql
celinepelletier: No it was a very good call, I'm wondering what the schema in the ETL doesn't enforce more columns

LysianeBouchard: Hum ... It look a bit strange indeed.  I propose to do the following:

1. Apply your suggestions, plus, if any, the fields already marked as required in the struct
2. Inspect the table in the CQDG QA environment to confirm none of those columns currently hold NULL values.
3. Open a separate PR in radiant-portal-pipeline to mirror the changes, and let it in draft mode until Karl is back to review/merge.

Does that work for you?

LysianeBouchard: Actually, I think maybe I should just use the schema that we currently have. I can probably obtain it in QA / CHOP.  If we want to modify the schema to add non null constraints and apply the change in QA/CHOP, we should probably have a dedicated ticket. Open to discuss this of course ...

celinepelletier: yeah take what comes from ETL and maybe open a Jira about that so we don't forget about it please

celinepelletier (approved):

celinepelletier (commented):
Don't forget to fix the conflicts

PR #1340, opened 2026-05-29
fix: SJRA-1529 fix leak between cases sharing same seq exp

celinepelletier:
# fix(variants): SJRA-1529 prevent occurrence leak across cases sharing a sequencing experiment

## 📌 Context

When two cases share a sequencing experiment, variant pages were showing occurrences from one case while viewing the other. The repo queries were joining `germline__snv__occurrence` to `case_has_sequencing_experiment` on `seq_id` alone, so any occurrence at a shared `seq_id` matched every case linked to that sequencing — regardless of which annotation task actually produced it.

The fix binds occurrences to a case via `task_context`'s `(case_id, seq_id, task_id)` triple, so each case only sees the occurrences produced by its own annotation task.

## 📝 Changes

**Variants leak fix (`backend/internal/repository/variants.go`)**
- `GetVariantInterpretedCases`: added a `task_context` bridge (`tctx.task_id = g_snv_o.task_id AND tctx.sequencing_experiment_id = g_snv_o.seq_id AND tctx.case_id = ig.case_id`) so the joined occurrence is scoped to the interpretation's case. `interpretation_germline` has no `task_id` column, hence the bridge.
- `GetVariantUninterpretedCases`: rewrote to drive from `task_context` and join the occurrence on both `seq_id` AND `task_id`. Conditional `JoinFamilyWithPatient` (only when `userQuery` filters on patient fields).
- `GetVariantCasesCount`: both counts rewritten as raw SQL with `COUNT(DISTINCT (case_id, seq_id, task_id))`. Interpreted side now bridges through `germline_snv_occurrence + task_context` to recover the missing `task_id`.

**Helper cleanup (`backend/internal/utils/repositories.go`)**
- Removed leak-prone helpers no longer used: `JoinGermlineSNVOccurrenceWithSeqExp`, `JoinGermlineSNVOccurrenceWithCaseHasSeqExp`, `JoinCaseHasSeqExpWithGermlineSNVOccurrence`, `AntiJoinCaseHasSeqExpWithGermlineInterpretationForLocus`.

**Dead-code removal — `GET /sequencing/:seq_id`**
- The endpoint had no frontend consumer (only `GET /sequencing/:seq_id/details` is used). Removed handler, repository, mock, integration test, route registration, security-test entry, and regenerated swagger / TypeScript / Python clients.

**Default sort + new field**
- `TaskContextTaskIdField` added under `TaskContextTable` and included in `VariantUninterpretedCasesDefaultFields` + the default sort, so each (case, seq) row carries its `task_id` and sorts deterministically when multiple annotation tasks exist.

**Test fixture realignment**
- Renumbered germline SNV `task_id`s in StarRocks fixtures (`simple`, `multiple`, `aggregation`, `pagination`) to reference real `radiant_germline_annotation` task IDs from the Postgres seed.
- Added matching case-scoped `task_context` rows in `clinical/05_task_context.sql` and `radiant_germline_annotation` rows in `clinical/01_task.sql`.
- Updated `clinical/05_occurrence_note.sql` to point notes at the new annotation `task_id`s.

## ☑️ TO-DOs

- [ ] Review

## 🧪 Tests

- `go test ./...` is green across the backend.
- `Test_GetVariantCasesCount` updated to assert `COUNT(DISTINCT (case, seq, task))` (interpreted=2, uninterpreted=4 at locus 1000).
- Existing variant tests (`Test_GetVariantInterpretedCases_*`, `Test_GetVariantUninterpretedCases_*`) cover the new join paths.
- Occurrence-leak repro tests (`Test_Germline_SNV_GetOccurrences_TaskIdScopesToOwningCase`, CNV/somatic equivalents) updated for the renumbered task IDs.

## 🔗 Related Issues

- [SJRA-1259](https://d3b.atlassian.net/browse/SJRA-1259)
- [SJRA-1529](https://d3b.atlassian.net/browse/SJRA-1529)

## 👀 Notes for Reviewers

- Three SQL queries changed shape; please sanity-check the `task_context` bridge in `GetVariantInterpretedCases` and the drive-from-`task_context` rewrite in `GetVariantUninterpretedCases`.
- Heads-up that the fixture renumbering touches a lot of files but the semantics are unchanged — task IDs in the StarRocks fixtures now correctly reference annotation tasks in the Postgres seed (which is what the task_context-driven queries require).
- `GetVariantCasesCount` was switched to raw SQL for readability; if you prefer it stays in the GORM builder for symmetry with the list functions, easy to flip back.
- Frontend `api.ts` regeneration drops the unused `getSequencing` method; no app code referenced it.

[SJRA-1529]: https://d3b.atlassian.net/browse/SJRA-1529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SJRA-1259]: https://d3b.atlassian.net/browse/SJRA-1259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/internal/repository/germline_cnv_occurrences_test.go
```diff
@@ -167,36 +167,36 @@ func Test_GermlineCNV_GetOccurrences_PaginationAndSorting(t *testing.T) {
...
-        // CNV3: nb_snv = 1 in DB -> expect 1
-        occurrences, err = repo.GetOccurrences(3, 7, 7, query)
-        occurrences, err = repo.GetOccurrences(3, 7, 7, query)
-        assert.NoError(t, err)
-        if assert.Len(t, occurrences, 1) {
-            assert.Equal(t, "CNV3", occurrences[0].Name)
-            assert.Equal(t, 1, occurrences[0].NbSNV)
-        }
-    })
+	testutils.ParallelTestWithStarrocks(t, "simple", func(t *testing.T, db *gorm.DB) {
```
LysianeBouchard, line 170: Il me semble que c'est moi qui avait ajouté ce test. De ce que je comprends, j'aurais accidentellement mis des espaces au lieu des tabs.

J'ajoute rouler gofmt à mon checklist de PR.

backend/internal/repository/germline_cnv_occurrences_test.go
```diff
@@ -167,36 +167,36 @@ func Test_GermlineCNV_GetOccurrences_PaginationAndSorting(t *testing.T) {
...
+		occurrences, err = repo.GetOccurrences(1, 2, 2, query)
+		assert.NoError(t, err)
+		if assert.Len(t, occurrences, 1) {
+			assert.Equal(t, "CNV2", occurrences[0].Name)
+			assert.Equal(t, 0, occurrences[0].NbSNV)
+		}
+
+		// CNV3: nb_snv = 1 in DB -> expect 1
+		occurrences, err = repo.GetOccurrences(3, 7, 7, query)
+		occurrences, err = repo.GetOccurrences(3, 7, 7, query)
```
LysianeBouchard: Pas ta faute, mais je vois que je crois que j'ai dupliqué la ligne ici. Si tu veux, tu peux corriger en même temps.

LysianeBouchard (approved):
💪

jecos (approved):

backend/internal/repository/variants.go
```diff
@@ -251,20 +266,44 @@ func (r *VariantsRepository) GetVariantCasesCount(locusId int) (*VariantCasesCou
...
 	locusIdString := fmt.Sprintf("%d", locusId)
-	txInterpreted := r.db.Table(fmt.Sprintf("%s %s", types.InterpretationGermlineTable.FederationName, types.InterpretationGermlineTable.Alias))
-	txInterpreted = txInterpreted.Select("COUNT(1)")
-	txInterpreted = txInterpreted.Where("locus_id = ?", locusIdString)
-	if err := txInterpreted.Find(&countInterpreted).Error; err != nil {
+
+	// Count distinct (case_id, seq_id, task_id) triples that have an
+	// interpretation at this locus. interpretation_germline has no task_id, so
+	// bridge through germline_snv_occurrence + task_context to recover it.
+	txInterpreted := r.db.Raw(`
```
kritchie, line 273: [nit] Just making sure, this change the style is intended using `r.db.Raw` instead of GORM-style queries ?

kritchie (approved):
Single question, non-blocking.
The rest LGTM ! 👌 🚀

backend/internal/repository/variants.go
celinepelletier: I changed for raw sql because it seemed more readable and there is no dynamic select/where/order clauses like for others queries. I can switch it back to gorm query builder if you prefer

kritchie: No preference, was just double checking 👍 
No changes required.

PR #1353, opened 2026-06-03
feat: SJRA-1452 install ranger in local deployment

jecos:
## 📌 Context

Adds **Apache Ranger** to the backend local `docker-compose` stack, following the direction of the
multi-tenancy ADR (`docs/adr/multi-tenancy-security.md`) and the Ranger POC (`docs/adr/ranger-poc/`).

Scope is intentionally limited to standing Ranger up and registering a StarRocks **service** in it.
There is **no StarRocks engine enforcement** yet (no `fe.conf` change / `access_control = ranger`),
so the existing API/worker behaviour is unchanged. Ranger is backed by a standalone Postgres that
**mimics the production external/managed DB**, so the install model matches what we'll do in prod.

## 📝 Changes

- Add two services to `backend/docker-compose.yml`:
  - `ranger` (`apache/ranger:2.7.0`) on port `6080`
  - `ranger-db` (`postgres:14`) — a plain Postgres standing in for the external/managed prod DB
- `backend/scripts/ranger/install.properties` — the Ranger DB connection, mounted over
  `/opt/ranger/admin/install.properties`. This is the single editable surface: for prod you change
  `db_host` + credentials to the managed endpoint and drop the local `ranger-db` container.
- `backend/scripts/ranger/create-ranger-services.py` — replaces the stock Ranger bootstrap script
  so the nine demo `dev_*` services are **not** created; instead it registers the StarRocks
  service-def and a single `starrocks` service (idempotent).
- `backend/scripts/ranger/ranger-servicedef-starrocks.json` — StarRocks Ranger service definition
  (catalog → database → table → column resources, access types, data-mask & row-filter defs).

## ☑️ TO-DOs

- [x] No outstanding tasks — change is self-contained and verified locally.

## 🧪 Tests

No automated tests (docker-compose infra). Verified manually:

```bash
cd backend && docker compose up -d ranger-db ranger
```

- `ranger` becomes healthy; Admin UI at http://localhost:6080 (login `admin` / `rangerR0cks!`).
- DB provisioned through the externalized `install.properties`:
  `ranger` database + `rangeradmin` role created in `ranger-db`.
- Service list shows **only** `starrocks` (plus Ranger's built-in `tag`) — none of the `dev_*` demo services.
- `starrocks` service-def is registered.

## 🔗 Related Issues

- [SJRA-1452](https://d3b.atlassian.net/browse/SJRA-1452)

## 👀 Notes for Reviewers

- **Why a separate Postgres** (not the app's `radiant` DB, not the `apache/ranger-db` image): to
  mirror an external/managed prod DB. The `apache/ranger:2.7.0` image bakes its DB config with no
  env override, which is why we mount a full `install.properties`.
- ⚠️ Ranger's property parser does **not** support trailing inline `#` comments on a value line
  (they get parsed into the value) — comments in `install.properties` are kept on their own lines.
- **"Test Connection" on the `starrocks` service fails** with *"Configuration validation is not
  implemented for starrocks"* — this is **expected** and harmless. It only needs the *optional*
  `ranger-starrocks-plugin` (autocomplete / test-connection only); policy enforcement does not
  depend on it, so we deliberately did not install it.
- A `tag` service also appears in Ranger — it's a built-in, not one of the demo services.

[SJRA-1452]: https://d3b.atlassian.net/browse/SJRA-1452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/scripts/ranger/create-ranger-services.py
+++ b/backend/scripts/ranger/create-ranger-services.py
@@ -1,0 +1,57 @@
+# Replaces the stock apache/ranger:2.7.0 create-ranger-services.py (which creates nine demo
+# dev_* service instances pointing at non-existent Hadoop/Hive/etc. containers).
+#
+# ranger.sh runs this once on first setup (when /opt/ranger/.setupDone is absent), ~30s after the
+# admin starts. By fully replacing the stock script, the dev_* services are never created; instead
+# we register our StarRocks service definition and a single "starrocks" service instance.
+#
+# Scope: service-def + service only. No access policies, and no StarRocks engine integration
+# (no fe.conf change) -- the service config below is used by Ranger only for test-connection /
+# resource lookup in the policy UI, and is harmless if StarRocks is unreachable.
+
+import json
+
+from apache_ranger.client.ranger_client import RangerClient
+from apache_ranger.model.ranger_service import RangerService
+from apache_ranger.model.ranger_service_def import RangerServiceDef
+
+ranger_client = RangerClient('http://ranger:6080', ('admin', 'rangerR0cks!'))
+
+SERVICEDEF_FILE = '/home/ranger/scripts/ranger-servicedef-starrocks.json'
+
+
+def missing(getter, name):
+    try:
+        return getter(name) is None
+    except Exception:
+        # The stock script treats a non-JSON / 404 response as "does not exist".
+        return True
+
+
+# 1. Register the StarRocks service definition (idempotent).
+with open(SERVICEDEF_FILE) as f:
+    starrocks_def = RangerServiceDef(json.load(f))
+
+try:
+    if missing(ranger_client.get_service_def, 'starrocks'):
+        ranger_client.create_service_def(starrocks_def)
+        print(" starrocks service-def created!")
+    else:
+        print(" starrocks service-def already exists")
+except Exception as e:
+    print(f"An exception occured creating the starrocks service-def: {e}")
+
+# 2. Create the StarRocks service instance (idempotent).
+starrocks = RangerService({'name': 'starrocks', 'type': 'starrocks',
+                           'configs': {'username': 'root', 'password': '',
+                                       'jdbc.driverClassName': 'com.mysql.cj.jdbc.Driver',
+                                       'jdbc.url': 'jdbc:mysql://starrocks:9030'}})
+
+try:
+    if missing(ranger_client.get_service, 'starrocks'):
+        ranger_client.create_service(starrocks)
+        print(" starrocks service created!")
+    else:
+        print(" starrocks service already exists")
+except Exception as e:
+    print(f"An exception occured creating the starrocks service: {e}")
--- a/backend/scripts/ranger/install.properties
+++ b/backend/scripts/ranger/install.properties
@@ -1,0 +1,112 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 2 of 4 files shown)


backend/scripts/ranger/install.properties
```diff
@@ -0,0 +1,112 @@
...
+keyadmin_password=rangerR0cks!
+
+
+# Audit shipping is a no-op without a Solr container; admin still boots fine (see backend/docker-compose.yml).
+audit_store=solr
+audit_solr_urls=http://ranger-solr:8983/solr/ranger_audits
+audit_solr_collection_name=ranger_audits
+
+# audit_store=elasticsearch
+audit_elasticsearch_urls=
```
celinepelletier, line 71: do we need to keep this ? we don't have any ES 🤔

celinepelletier (approved):

backend/scripts/ranger/create-ranger-services.py
```diff
@@ -0,0 +1,57 @@
...
+# (no fe.conf change) -- the service config below is used by Ranger only for test-connection /
+# resource lookup in the policy UI, and is harmless if StarRocks is unreachable.
+
+import json
+
+from apache_ranger.client.ranger_client import RangerClient
+from apache_ranger.model.ranger_service import RangerService
+from apache_ranger.model.ranger_service_def import RangerServiceDef
+
+ranger_client = RangerClient('http://ranger:6080', ('admin', 'rangerR0cks!'))
```
LysianeBouchard, line 18: If this script only runs locally via docker-compose, I assume hardcoding the username/password in cleartext like this is fine.

Is that the case?

backend/scripts/ranger/install.properties
```diff
@@ -0,0 +1,112 @@
...
+#
+# db_root_user / db_root_password: privileged bootstrap account; creates db_name + db_user and loads schema.
+db_root_user=postgres
+db_root_password=rangerR0cks!
+# db_host: in production, point this at the external/managed Postgres endpoint (host[:port]).
+db_host=ranger-db
+
+db_name=ranger
+# db_user / db_password: runtime account Ranger uses after bootstrap.
+db_user=rangeradmin
```
kritchie, line 50: Just making sure, I don't see this `rangeradmin` created elsewhere, maybe it's there by default? 🤔

kritchie (approved):
Small question, but most likely not blocking (due to my limited knowledge of Ranger)
Otherwise LGTM 👍

backend/scripts/ranger/create-ranger-services.py
```diff
@@ -0,0 +1,57 @@
...
+
+SERVICEDEF_FILE = '/home/ranger/scripts/ranger-servicedef-starrocks.json'
+
+
+def missing(getter, name):
+    try:
+        return getter(name) is None
+    except Exception:
+        # The stock script treats a non-JSON / 404 response as "does not exist".
```
LysianeBouchard, line 27: Claude told me that here you tried to mimic the previous behaviour of the stock script, but that actually, the exception catched in the stock script would be a JSONDecodeError.

Maybe he is just hallucinating things or looking at another version, but mentioning just in case.

backend/scripts/ranger/create-ranger-services.py
jecos: yes

backend/scripts/ranger/install.properties
jecos: ranger will create db and users

backend/scripts/ranger/install.properties
jecos: yes, it is not used but otherwise it does not start

PR #1359, opened 2026-06-04
feat: SJRA-1452 Ranger StarRocks image (build & push)

jecos:
# Ranger StarRocks image

## 📌 Context

The local deployment needs an Apache Ranger admin that authorizes StarRocks. The stock `apache/ranger:2.8.0` image is dev-only (always runs the full create-DB flow, exposes just 3 env vars, no StarRocks plugin). This PR replaces it with a **self-contained, dual-mode Ranger image** with the StarRocks integration baked in, wires it into `docker-compose.yml`, and publishes it from CI as `ranger-starrocks`.

## 📝 Changes

- **Custom Ranger image** (`backend/ranger/Dockerfile`, `FROM apache/ranger:2.8.0`):
  - Three entry-point modes (`backend/ranger/ranger-entrypoint.sh`): `migrate` (schema setup, exit — gated k8s Job), `serve` (setup + start admin + register), `register` (REST-only against a running admin).
  - `install.properties.tpl` — `setup_mode=SeparateDBA` (managed Postgres, no runtime superuser), all secrets via `RANGER_*` env (no plaintext committed).
  - `register-services.py` — generic file-driven registration (service-defs create-or-update, instances create-only; `${RANGER_*}` placeholders substituted before parse).
  - **StarRocks baked in**: plugin jar + MySQL JDBC driver downloaded onto the admin classpath; service-def (`ranger-servicedef-starrocks.json`) + instance (`starrocks-service.json`) shipped in the image.
- **Two runtime knobs** (resolved in the entry point):
  - `RANGER_STARROCKS_AUTOCOMPLETE` (default `yes`) → sets the service-def `implClass` (UI Test Connection / resource autocomplete); any other value leaves it empty and Ranger never connects to StarRocks — connection props become optional.
  - `RANGER_STARROCKS_SERVICE_NAME` (default `starrocks`) → registered service-instance name.
- **`backend/docker-compose.yml`**: `ranger` service now builds the custom image; added `ranger-db` (pre-creates DB + role for SeparateDBA). Removed the old plugin-download sidecar, named volume, and JSON bind mounts (now baked).
- **`backend/ranger/README.md`**: documents the image, modes, full `RANGER_*` env contract, and that the StarRocks connection is optional when autocomplete is off.
- **`.github/workflows/build_and_push.yml`**: new `ranger-starrocks` image job (metadata + build/push + provenance attestation), publishing to `ghcr.io/radiant-network/ranger-starrocks` on `main` / `v*`.

## ☑️ TO-DOs

- [ ] Confirm `RANGER_STARROCKS_SERVICE_NAME: starrocks_test` in `docker-compose.yml` is the intended local value (vs. default `starrocks`).

## 🧪 Tests

Verified locally with `docker compose`:

- `docker compose build ranger` succeeds; plugin jars present in the image, service JSON baked in, entry point executable.
- Knob resolution: default → `implClass` set + name `starrocks`; `AUTOCOMPLETE=no` → `implClass` empty; custom `RANGER_STARROCKS_SERVICE_NAME` honored.
- `docker compose up ranger-db starrocks ranger` → admin healthy; `serve` registers the service-def + instance from the baked files.
- Test Connection: `POST /service/plugins/services/validateConfig` → `{"statusCode":0,"msgDesc":"Connection test successful"}`.
- `.github/workflows/build_and_push.yml` validated as well-formed YAML.

## 🔗 Related Issues

- [SJRA-1452](https://d3b.atlassian.net/browse/SJRA-1452)

## 👀 Notes for Reviewers

- `docker-compose.yml` includes some incidental formatting-only normalization (quote/indent) on unrelated services (`keycloak`, `api`, …) from an editor reformat.
- The image is generic by design: StarRocks is one possible mounted/baked service-def. The same entry point supports `migrate`/`register` Jobs for k8s.
- Replaces the now-closed PR #1353 for the same branch.

[SJRA-1452]: https://d3b.atlassian.net/browse/SJRA-1452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/ranger/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+COPY ranger-entrypoint.sh             /home/ranger/scripts/ranger-entrypoint.sh
+COPY register-services.py             /home/ranger/scripts/register-services.py
+
+# StarRocks service-def (type) + service instance, baked into the dirs register-services.py scans.
+COPY ranger-servicedef-starrocks.json /ranger/service-defs/starrocks.json
+COPY starrocks-service.json           /ranger/services/starrocks.json
+
+# StarRocks Ranger plugin jars on the admin classpath, so test-connection / autocomplete work at first
+# boot.
+ARG STARROCKS_PLUGIN_JAR=https://releases.starrocks.io/resources/ranger-starrocks-plugin-3.0.0-SNAPSHOT.jar
```
LysianeBouchard, line 34: It's normal for the `SNAPSHOT` keyword in the artifact name?

backend/ranger/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+RUN PLUGIN_DIR=/opt/ranger/admin/ews/webapp/WEB-INF/classes/ranger-plugins/starrocks && \
+    mkdir -p "${PLUGIN_DIR}" && \
+    curl -fsSL -o "${PLUGIN_DIR}/ranger-starrocks-plugin.jar" "${STARROCKS_PLUGIN_JAR}" && \
+    curl -fsSL -o "${PLUGIN_DIR}/mysql-connector-j.jar"       "${MYSQL_CONNECTOR_JAR}" && \
+    chmod +x /home/ranger/scripts/ranger-entrypoint.sh && \
+    chown -R ranger:ranger "${PLUGIN_DIR}" /home/ranger/scripts /ranger
+USER ranger
+
+ENV RANGER_STARROCKS_SERVICE_NAME=starrocks
+ENV RANGER_STARROCKS_AUTOCOMPLETE=yes
```
LysianeBouchard: Claude made me noticed that the default values for these variables are defined twice, here and also in ranger-entrypoint.sh.  If someone update, it must remember to update in both places. Maybe would be enough to only define the variables here and not i nthe script.

It's a very minor detail.  Consider it optional,  leaving the final decision to you.

backend/ranger/Dockerfile
jecos: yes, it comes from the official starrocks documentation
https://docs.starrocks.io/docs/3.5/administration/user_privs/authorization/ranger_plugin/#optional-install-ranger-starrocks-plugin

backend/ranger/Dockerfile
jecos: fixed

backend/ranger/install.properties.tpl
```diff
@@ -0,0 +1,130 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+# install.properties TEMPLATE for the dual-mode production Ranger image.
+#
+# ranger-entrypoint.sh renders this file to /opt/ranger/admin/install.properties at runtime, expanding
+# the ${...} placeholders below from the container environment (python3 os.path.expandvars). setup.sh
```
LysianeBouchard: This comment mentions os.path.expandvars, but the entrypoint actually uses string.Template.safe_substitute. Might be simplest to just drop the technical precision here and let ranger-entrypoint.sh speak for itself. Small detail though :)

backend/ranger/install.properties.tpl
```diff
@@ -0,0 +1,130 @@
...
+# audit_store=elasticsearch
+audit_elasticsearch_urls=
+audit_elasticsearch_port=9200
+audit_elasticsearch_protocol=http
+audit_elasticsearch_user=elastic
+audit_elasticsearch_password=elasticsearch
+audit_elasticsearch_index=ranger_audits
+audit_elasticsearch_bootstrap_enabled=true
+
+policymgr_external_url=http://ranger-admin:6080
```
LysianeBouchard: Si c'est un composant ranger qui doit être accédé par d'autres service, pourrait être pertinent de paramétrer ... Je te laisse valider.

backend/ranger/ranger-entrypoint.sh
```diff
@@ -0,0 +1,126 @@
...
+
+#
+# Dual-mode entrypoint for the production Apache Ranger image.
+#
+#   migrate  render install.properties from env, run setup.sh (schema only), exit 0.
+#            Use as a gated/run-once k8s Job before rolling out serve pods.
+#   serve    render install.properties, run setup.sh (idempotent; also regenerates
+#            ranger-admin-site.xml in this pod's filesystem), start the admin, then (unless
+#            RANGER_REGISTER_ON_SERVE=false) apply any mounted service-defs/services, and stay alive.
+#   register apply mounted service-def / service JSON to a RUNNING admin (no DB setup, no server).
```
LysianeBouchard: Very small detail, but I think it is not longer mounted. You might want to adjust a bit.

kritchie (approved):

backend/ranger/register-services.py
```diff
@@ -0,0 +1,101 @@
...
+
+def load_json(path):
+    with open(path) as fh:
+        return json.loads(Template(fh.read()).safe_substitute(_subst))
+
+
+def missing(getter, name):
+    try:
+        return getter(name) is None
+    except Exception:
```
LysianeBouchard, line 67: It's a bit loose here.  You might want something a bit more precise if it is for production use ... Leave it to your judgement.

backend/ranger/install.properties.tpl
jecos: no it's not ok, good catch

backend/ranger/register-services.py
jecos: i think this one is ok

PR #1364, opened 2026-06-04
feat: SJRA-1445 tenant path-prefix routing — first slice (HPO autocomplete + enforcement toggle)

celinepelletier:
## 📌 Context

First slice of multi-tenancy tenant routing (ADR §6 — "REST API tenant routing, path prefix"). It establishes the `/{tenant}/…` path-prefix pattern and the tenant-membership middleware end-to-end on a single low-risk endpoint (`GET /hpo/autocomplete`), behind a default-off toggle so it can ship before users are backfilled into `user_role`.

`/hpo/autocomplete` was chosen for the first slice because it has **zero frontend callers and zero Cypress references** — minimal blast radius while still exercising the full pipeline (route group, middleware + toggle, `HasTenantAccess`, swagger, client regen, Postman). Migrating the remaining tenant-scoped endpoints is a follow-up PR.

## 📝 Changes

- **`RequireTenantAccess` middleware** (`internal/server/middlewares.go`): reads the `:tenant` path param, stores it in request context (`GetTenant` helper); when enforcement is on, verifies the caller holds ≥1 role in the tenant and returns **403** on cross-tenant access.
- **`TENANT_ENFORCEMENT_ENABLED` toggle** (default **off**), wired in `cmd/api/main.go`. With it off, routing + context still work but nobody is locked out; flip it on after the `user_role` backfill.
- **`HasTenantAccess(email, tenantCode)`** added to `AuthRepository` / `AuthRepositoryDAO`.
- **`GET /hpo/autocomplete` moved under `/{tenant}/`** (the only route relocated this PR). `/auth/*`, `/users/*`, `/status` and all other routes stay at root.
- Swagger `tenant` path param on `GetHPOTermAutoComplete` + regenerated `swagger.{yaml,json}`; regenerated **TS + Python** clients (only `hpoTermAutoComplete` gains the `tenant` arg — no frontend call-site changes, since it has no callers).
- **Postman**: added a `tenant` collection variable (`radiant`) and prefixed the HPO autocomplete request.
- Documented `TENANT_ENFORCEMENT_ENABLED` in `backend/CLAUDE.md`.

## ☑️ TO-DOs

- [ ] Follow-up PR: migrate the remaining tenant-scoped endpoints under `/{tenant}/`.
- [ ] Backfill users into `user_role`, then set `TENANT_ENFORCEMENT_ENABLED=true`.
- [ ] Update the Cypress suite to the new URLs once deployed.

## 🧪 Tests

- Full test suite is green

## 🔗 Related Issues

- [SJRA-1445](https://d3b.atlassian.net/browse/SJRA-1445)

## 👀 Notes for Reviewers

- **Toggle defaults OFF intentionally** — enforcing before `user_role` is backfilled would 403 every user. Trade-off: the check stays disabled until someone flips the env var (tracked in TO-DOs).
- Gin gives static routes priority over the `:tenant` wildcard, so `/auth`, `/users`, `/status` are never captured as a tenant.
- The middleware always resolves + stores the tenant regardless of the toggle, so handlers can start consuming `GetTenant` in later work without depending on enforcement being on.

[SJRA-1445]: https://d3b.atlassian.net/browse/SJRA-1445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/cmd/api/main.go
```diff
@@ -93,9 +93,18 @@ func setupRouter(dbStarrocks *gorm.DB, dbPostgres *gorm.DB) *gin.Engine {
...
 
+	// Global routes (not tenant-scoped) stay on privateRoutes: /auth/*, /users/*.
 	authGroup := privateRoutes.Group("/auth")
 	authGroup.GET("/me", server.GetMeHandler(repoAuth, auth))
 
+	// Tenant-scoped routes live under /:tenant and require the caller to hold at least one
+	// role in that tenant (cross-tenant access → 403). The resolved tenant is stored in context.
+	// Enforcement is gated by TENANT_ENFORCEMENT_ENABLED (default off) so routing can ship
+	// before users are backfilled into user_role; flip it on once the backfill lands.
+	enforceTenantAccess := utils.GetEnvOrDefault("TENANT_ENFORCEMENT_ENABLED", "false") == "true"
```
LysianeBouchard: Hum ... je me demande s'il n'y aurait pas un utilitaire go qui serait plus robuste ici?
Claude me recommande strconv.ParseBool

backend/internal/server/handlers.go
```diff
@@ -94,12 +94,13 @@ func GetMondoTermAutoComplete(repo repository.TermsDAO) gin.HandlerFunc {
 // @Description Retrieve AutoCompleteTerm list of HPO terms matching input string with highlighted
 // @Tags hpo
 // @Security bearerauth
+// @Param tenant path string true "Tenant code"
 // @Param prefix query string true "Prefix"
 // @Param limit query string false "Limit"
 // @Produce json
 // @Success 200 {array} types.AutoCompleteTerm
 // @Failure 500 {object} types.ApiError
```
LysianeBouchard, line 104: I think maybe we can have a 403 now if the feature toggle is activated?

backend/internal/repository/auth.go
```diff
@@ -48,6 +49,22 @@ func (r *AuthRepository) HasAction(email, tenantCode, orgCode, actionCode string
...
+// caller a member of the tenant.
+func (r *AuthRepository) HasTenantAccess(email, tenantCode string) (bool, error) {
+	var allowed bool
+	err := r.db.Raw(`
+		SELECT EXISTS (
+			SELECT 1 FROM user_role ur
+			WHERE ur.email = ? AND ur.tenant_code = ?
+		)`, email, tenantCode).Scan(&allowed).Error
+	if err != nil {
+		return false, fmt.Errorf("error checking tenant access for %q in %q: %w", email, tenantCode, err)
```
LysianeBouchard, line 63: My claude recommend to make the email comparison case insensitive. It's not a bad idea I think, but I guess it depends of the flow with the email ... What do you think?

backend/internal/repository/auth_test.go
LysianeBouchard, line 1: Could add a bit more protection to include users with different org column value in user_role. Ex: user with a single org-scope role, user with * org role, etc.  Leaving to your judgement; ok with me if you think it is too much corner cases and tests.

jecos (approved):

kritchie (approved):
Nice 👌

backend/internal/repository/auth_test.go
celinepelletier: added

backend/internal/repository/auth.go
celinepelletier: `LOWER(ur.email) = LOWER(?)` in each query defeats the email primary-key index and has to be remembered in every future query. Better to normalize at the boundary — lowercase the email when reading it from the token (and store it lowercased in the backfill / users table), keeping the index usable and the invariant in one place. That makes it a small normalization decision coupled to the backfill rather than a per-query patch.
What do you think about this approach?

LysianeBouchard (approved):
Big thanks for addressing my comments :) .

backend/internal/repository/auth.go
celinepelletier: I just added the lowercase on email when retrieving it from the token, the rest will be done when doing users migration and users creation

PR #1368, opened 2026-06-05
feat: CLIN-5997 PATCH cases with Tasks

creativeyann17:
- [x] Add `Tasks` to PATCH `/cases/batch`
- [x] Unit tests
- [x] Local end to end test

Required for: https://github.com/Ferlab-Ste-Justine/clin-qlin-me-hybrid/pull/351


backend/docs/swagger.yaml
kritchie, line 1: I think there's some updates that were not captured. 
Running the following should yield new updates:
- In `backend/` : `make doc`
- At root : `make generate-client-all`

backend/cmd/worker/case_patch_validation.go
```diff
@@ -12,13 +12,15 @@ import (
...
-// already-existing sequencing experiments to an already-existing case.
+// already-existing sequencing experiments to an already-existing case, and
+// optionally attaches bioinformatic tasks (with their documents) to it.
 type PatchCaseValidationRecord struct {
 	batchval.BaseValidationRecord
 	Patch                  types.CaseBatchPatch
 	CaseID                 *int
 	SequencingExperiments  map[int]*types.SequencingExperiment
 	DiagnosisLabCodeUpdate string
+	TaskCR                 *CaseValidationRecord
```
celinepelletier: This line is confusing regarding naming, is it a task ? or a case ?

backend/docs/swagger.yaml
creativeyann17: `make generate-client-all` update/add tones of files, some have nothing to do with my past PRs is that ok ?

kritchie: Yes, those are auto-generated client files (for Typescript and Python), they need to be kept in sync with the repo. It adds some noise in the PRs but it's necessary. 

In an ideal world this would be automated so we don't have to worry about this when opening a PR, but it's not currently the case.

creativeyann17: i would prefer to open another PR honeslty cause it's gonna make that one un-readable :)

celinepelletier: The generated code is well localized (api.ts and python/*) and it's better to have all the changes for a given scope in the same PR

backend/cmd/worker/case_validation_patch_integration_test.go
```diff
@@ -0,0 +1,191 @@
...
+	createDocumentsForBatch(ctx, client, base)
+	baseBytes, _ := json.Marshal(base)
+	id := insertPayloadAndProcessBatch(db, string(baseBytes), types.BatchStatusPending, types.CaseBatchType, false, "user123", "2025-12-04")
+	assertBatchProcessing(t, db, id, types.BatchStatusSuccess, false, "user123", emptyMsgs, emptyMsgs, emptyMsgs)
+}
+
+// Happy path: a PATCH that carries a sequencing experiment + a new task persists
+// the task, its output document and the task_has_document link.
+func Test_ProcessBatch_PatchCase_AttachTask_Success(t *testing.T) {
+	testutils.SequentialTestWithPostgresAndMinIO(t, func(t *testing.T, ctx context.Context, client *minio.Client, endpoint string, db *gorm.DB) {
```
celinepelletier: Use the Need framework instead of the old test functions

backend/cmd/worker/case_validation_patch_integration_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+// PATCH appends tasks; it does NOT dedup them. The base case already carries an
+// alignment_germline_variant_calling task on NA12891 (from the POST seed); a PATCH that
+// re-sends that experiment plus a same-type task on the same aliquot ends up with TWO such
+// tasks (persistTask always CreateTask()s, there is no UNIQUE on document.url, and the
+// task_context/task_has_document keys both include the fresh task_id). Two things are
+// asserted here, and they pull in opposite directions on purpose:
+//   - the experiment attachment IS idempotent (case_has_sequencing_experiment is ON CONFLICT
+//     DO NOTHING), so re-sending the experiment adds no row;
+//   - the task is NOT idempotent — it is appended, duplicating type_code + aliquot.
```
celinepelletier, line 140: So the patch is an append only for task, would you have a need to delete a task in the patch use case ?

backend/cmd/worker/case_patch_validation.go
```diff
@@ -12,13 +12,15 @@ import (
...
-// already-existing sequencing experiments to an already-existing case.
+// already-existing sequencing experiments to an already-existing case, and
+// optionally attaches bioinformatic tasks (with their documents) to it.
 type PatchCaseValidationRecord struct {
 	batchval.BaseValidationRecord
 	Patch                  types.CaseBatchPatch
 	CaseID                 *int
 	SequencingExperiments  map[int]*types.SequencingExperiment
 	DiagnosisLabCodeUpdate string
+	Task                 *CaseValidationRecord
```
celinepelletier: What happen if we want to add 2 tasks on an existing case ?

creativeyann17: append :)

backend/cmd/worker/case_validation_patch_integration_test.go
creativeyann17: we dont delete, same for sequencing experiement, for delete we could add a query param saying delete/replace i dont know what's the standard, default behavior is append to existing

backend/cmd/worker/case_patch_validation.go
```diff
@@ -91,9 +96,77 @@ func validatePatchCaseRecord(cache *batchval.BatchValidationCache, patch types.C
...
+	if len(patch.Tasks) > 0 {
+		if err := validatePatchTasks(ctx, cache, patch, index, r); err != nil {
+			return nil, err
+		}
+	}
+
 	return r, nil
 }
 
+// validatePatchTasks reuses the POST /cases/batch task machinery to validate the patch's
```
kritchie, line 108: [very-nit] Maybe the comments can be scoped from the `worker`'s perspective ? Instead of refering to `POST /cases/batch` endpoint, we mention the specific component of the worker, so here the `CaseValidationRecord`. Because technically the worker doesn't need to API to exist, we could write batches manually in the DB and it would still be valid.

backend/cmd/worker/case_patch_validation.go
```diff
@@ -91,9 +96,77 @@ func validatePatchCaseRecord(cache *batchval.BatchValidationCache, patch types.C
...
+	taskRecord := NewCaseValidationRecord(ctx, cache, taskCase, index)
+	taskRecord.CaseID = r.CaseID
+
+	if err := taskRecord.fetchTaskTypeCodes(); err != nil {
+		return fmt.Errorf("fetch task type codes: %w", err)
+	}
+	if err := taskRecord.fetchDocumentCodes(); err != nil {
+		return fmt.Errorf("fetch document codes: %w", err)
+	}
+	// Resolves sequencing experiments (by task aliquot), their existing task contexts, and
```
kritchie, line 164: Do we check if the sequencing experiments fetched from the tasks match those that are associated to the case ? 🤔 (I probably missed that)

backend/cmd/worker/case_patch_validation.go
celinepelletier: I'm lost sorry, for sequencing we have a collection map[int]*types.SequencingExperiment to have multiple sequencing for a given line of the batch but for case I don't see a collection of task so I don't understand how it will work 🤔

celinepelletier: Oh await the tasks are in Patch, so this Task property is used to validate the tasks but doens't include them rigth?

backend/docs/swagger.yaml
creativeyann17: my bad installed the wrong version of the doc generator

backend/cmd/worker/case_patch_validation.go
creativeyann17: Yes only with the `sequencing_experiments` from the same payload which is fine and logical from QLIN point of view cause we send both at the same time BUT maybe from radiant point of view we should check also in DB ... i can add that

backend/cmd/worker/case_patch_validation.go
```diff
@@ -91,9 +96,77 @@ func validatePatchCaseRecord(cache *batchval.BatchValidationCache, patch types.C
...
+	taskCase := types.CaseBatch{
+		SubmitterCaseId:       patch.SubmitterCaseId,
+		ProjectCode:           patch.ProjectCode,
+		SequencingExperiments: seqExps,
+		Tasks:                 tasks,
+	}
+	taskRecord := NewCaseValidationRecord(ctx, cache, taskCase, index)
+	taskRecord.CaseID = r.CaseID
+
+	if err := taskRecord.fetchTaskTypeCodes(); err != nil {
```
kritchie, line 158: Probably something we missed in the original implementation, but it looks like we allow setting any type of task for an existing case. We might introduce some inconsistencies here with regards to `germline` <-> `somatic` cases and tasks combination. 

Not sure if this is something we want to address now ? 🤔

celinepelletier (approved):

kritchie (approved):

PR #1371, opened 2026-06-08
feat: SJRA-1445 move others route under tenant and add default tenant in frontend

celinepelletier:
# feat: SJRA-1445 migrate all tenant-scoped routes under /{tenant} prefix

## 📌 Context

Follow-up to the first slice (#1364), which established the `/{tenant}/…` path-prefix pattern, the `RequireTenantAccess` middleware, and the `TENANT_ENFORCEMENT_ENABLED` toggle on a single endpoint (`/hpo/autocomplete`). This PR completes the migration: **every remaining tenant-scoped endpoint** now lives under `/{tenant}/…` (ADR §6 — "REST API tenant routing, path prefix"). Global endpoints (`/auth/*`, `/users/*`, `/status`) stay at root.

Enforcement still defaults **off** (`TENANT_ENFORCEMENT_ENABLED`), so this can deploy before users are backfilled into `user_role` — routing and request-context work with no lockout; flip the flag on after the backfill.

## 📝 Changes

- **Routing** (`cmd/api/main.go`): all tenant-scoped groups (`cases`, `genes`, `hpo`, `igv`, `interpretations`, `mondo`, `notes`, `occurrences`, `sequencing`, `variants`, `documents`, and the `data_manager` batch routes) moved under `tenantRoutes := privateRoutes.Group("/:tenant")`. The `data_manager` block stays nested under it, so batch routes layer JWT → tenant membership → `data_manager` role.
- **Swagger**: added a `tenant` path param and `@Failure 401` / `@Failure 403` to all 68 remaining tenant-scoped handlers (sorted into existing failure lists; batch handlers already documented `403` for the role check). Regenerated `docs/swagger.{yaml,json}`.
- **Clients**: regenerated TS (`frontend/api`) and Python (`cli/python`) — `tenant` is the leading argument on every tenant-scoped method.
- **Frontend**: every tenant-scoped client call passes the shared `DEFAULT_TENANT = 'radiant'` constant (used across 44 files). Temporary bridge until the tenant switcher ships; global (`/users/*`) calls unchanged. @AltefrohneGaelle @lflangis 
- **Postman**: all 65 tenant-scoped requests prefixed with `{{tenant}}`.
- **Tests**: every tenant-scoped handler unit test and `cmd/api` integration test now registers `/:tenant/...` and requests `/radiant/...` to mirror the real mounting; `Test_SecureRoutes` updated for all moved routes. Globals left at root.

## ☑️ TO-DOs

- [ ] Backfill users into `user_role`, then set `TENANT_ENFORCEMENT_ENABLED=true`.
- [ ] Update the Cypress suite to the `/{tenant}/` URLs (deferred — runs against live QA, not yet deployed).
- [ ] Have handlers consume the resolved tenant (`GetTenant`) to scope data — separate task.

## 🧪 Tests

- `go test ./internal/server/` — handler unit tests pass (now exercise `/:tenant/...` registration + `/radiant/...` requests).
- `go test ./cmd/api/` — full integration suite passes (~103s), incl. `Test_SecureRoutes` against the real router for every moved route.
- `go build ./...` and `go vet ./...` clean (only pre-existing unkeyed-literal warnings).
- Frontend: no new `TS2554` arity errors after the client regen + call-site patch.
- Tested locally with enforcement ON and OFF

## 🔗 Related Issues

- [SJRA-1445](https://d3b.atlassian.net/browse/SJRA-1445)

## 👀 Notes for Reviewers

- **Enforcement defaults OFF**, so the `403` advertised on every route won't fire until the flag is enabled post-backfill. The spec documents the contract, not the current flag state (same convention as the first slice).
- Gin gives static routes priority over the `:tenant` wildcard, so `/auth`, `/users`, `/status` are never captured as a tenant (the names are effectively reserved).
- Large diff is mostly generated artifacts (`frontend/api/api.ts`, `cli/python/*`). The hand-written changes are `main.go`, the handler Swagger annotations, and the test path updates.
- `frontend` does not `tsc` fully clean due to ~560 pre-existing `TS1484` (verbatimModuleSyntax) errors unrelated to this PR; the bar here is "no new arity errors," which holds.
- Batch/ingestion routes are now tenant-scoped but don't yet *consume* the tenant — intentionally deferred, like the other handlers.

[SJRA-1445]: https://d3b.atlassian.net/browse/SJRA-1445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AltefrohneGaelle (approved):
Front part only

kritchie (approved):

PR #1373, opened 2026-06-08
feat: SJRA-1474 StarRocks SSL + Keycloak JWT auth with Ranger authorization, and a createuser provisioning tool

jecos:
# feat: SJRA-1474 StarRocks SSL + Keycloak JWT auth with Ranger authorization, and a `createuser` provisioning tool

## 📌 Context

StarRocks authentication and authorization are being moved onto the platform's identity stack: Keycloak becomes the authentication provider (JWT) and Apache Ranger becomes the **sole** authorization authority (replacing StarRocks' built-in grants). This PR enables TLS on StarRocks, wires the JWT/Ranger trust, and adds a Go tool that provisions a user end-to-end across all four backing systems — keyed on the immutable Keycloak `sub` so `current_user()` in StarRocks resolves straight through to the auth model. The provisioning code is structured so it can later sit behind a real `POST /users` handler, not just the local seeding command.

## 📝 Changes

- **StarRocks security**: enable SSL (FE config + self-signed cert generation for the compose stack); authenticate via Keycloak JWT (`authentication_jwt`, `principal_field = sub`) with an added **audience** check.
- **Ranger as sole authority**: access enforced on tables, masking on columns; no StarRocks built-in grants. `03_ranger_policies.py` creates tenant roles empty + create-if-missing so re-runs don't wipe members the provisioning tool added.
- **Ranger move (SJRA-1472)**: Ranger runs from the published image via `backend/compose/ranger-compose.yml` (3 phases: migrate / serve / register), included by `backend/compose/docker-compose.yml`; removed the in-tree `backend/ranger/` image sources.
- **Compose reorganization**: relocated `docker-compose.yml` under `backend/compose/` and added setup scripts (seed Postgres, StarRocks users/views, Ranger policies, an end-to-end `04_verify.py`), cert generation, `fe.conf`, and Ranger security XML.
- **`createuser` tool** (`backend/cmd/createuser`): provisions one regular user across Keycloak → Postgres → Ranger → StarRocks, all keyed on the `sub`. Idempotent; seeds alice/bob/wendy with no flags, or a single ad-hoc user via flags.
- **Package layout** (split by concern for reuse behind a future API):
  - `internal/service/admin.go` — `ProvisionUser` orchestrator + the consumer-side interfaces it depends on.
  - `internal/client/admin_keycloak.go`, `admin_ranger.go` — Keycloak/Ranger admin REST clients (stdlib net/http). Keycloak `password` is optional: empty skips the credential reset, for realms delegating to an external IdP.
  - `internal/repository/` — `AuthRepository` write side (`UpsertUser`/`GrantRole`); `StarrocksUserRepository` for the JWT-user DDL.
  - `internal/types/user_provisioning.go` — shared `Grant`/`UserInput`.
- **Docs/rules**: added `backend/.claude/rules/go-code-review.md` (project-annotated Go Code Review Comments, incl. the consumer-side interface-placement convention), linked from `CLAUDE.md` and `universal.md`.
- Merged `main` and resolved the docker-compose conflict (main had merged the SJRA-1472 ranger move independently in #1367).

## ☑️ TO-DOs

- [ ] None blocking.

## 🧪 Tests

- `go build ./...` and `go vet` clean on touched packages (verified after merging `main`).
- Unit tests (no containers): `internal/client` (httptest fakes for Keycloak/Ranger), `internal/service` (mock-based orchestrator), `cmd/createuser` (flag parsing + demo-grant drift guard).
- Integration tests against testcontainer Postgres: `AuthRepository` write side; plus the StarRocks DDL helper unit test.
- `docker compose -f backend/compose/docker-compose.yml config` validates (the `ranger-compose.yml` include resolves).
- Compose stack validated end-to-end via `04_verify.py` (PII masking checks pass).

Run:
```bash
cd backend
go build ./... && go vet ./...
go test ./internal/client/... ./internal/service/... ./cmd/createuser/...
go test ./internal/repository/...   # needs Docker (testcontainers)
```

## 🔗 Related Issues

- [SJRA-1474](https://d3b.atlassian.net/browse/SJRA-1474)

## 👀 Notes for Reviewers

- The interesting design choice is interface placement: `service` defines the provisioner interfaces it consumes (consumer-side) and `client`/`repository` return concrete structs — so `service` imports only `internal/types`. Rationale and the heuristic are written up in `backend/.claude/rules/go-code-review.md`.
- The `sub` is the single identity key across all four systems (StarRocks username = `users.user_id` = Ranger user name). Every provisioning step is idempotent, so re-running converges.
- `AuthRepository`'s write methods were intentionally **not** added to `AuthRepositoryDAO` (that interface is the read-only handler contract; adding writes would break the `GetMeHandler` mock). The service's `AuthStore` interface is satisfied by `*AuthRepository` directly.

[SJRA-1472]: https://d3b.atlassian.net/browse/SJRA-1472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SJRA-1474]: https://d3b.atlassian.net/browse/SJRA-1474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/cmd/createuser/main.go
```diff
@@ -0,0 +1,159 @@
...
+	return nil
+}
+
+func main() {
+	var (
+		email    = flag.String("email", "", "user email (PK). If empty, the demo users alice/bob/wendy are provisioned.")
+		username = flag.String("username", "", "Keycloak username (defaults to the local part of -email)")
+		first    = flag.String("first", "", "first name")
+		last     = flag.String("last", "", "last name")
+		password = flag.String("password", utils.GetEnvOrDefault("USER_PASSWORD", "radiant123!"), "user password")
```
celinepelletier: Isn't it dangerous to have a default value for PASSWORD?

backend/cmd/createuser/main.go
```diff
@@ -0,0 +1,159 @@
...
+			name = strings.SplitN(*email, "@", 2)[0]
+		}
+		users = []types.UserInput{{
+			Username: name, Email: *email, FirstName: *first, LastName: *last,
+			Password: *password, Grants: grants,
+		}}
+	} else {
+		// Demo users share the configured password.
+		for i := range users {
+			users[i].Password = *password
```
celinepelletier: Is it safe to give the same password at multiple users ?

backend/cmd/createuser/main.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	return service.AdminDeps{
+		Keycloak:  client.NewKeycloakAdminClient(client.KeycloakConfigFromEnv()),
+		Ranger:    client.NewRangerAdminClient(client.RangerConfigFromEnv()),
+		Starrocks: repository.NewStarrocksUserRepository(sr, repository.StarrocksJWTConfigFromEnv()),
+		Auth:      repository.NewAuthRepository(pg),
+	}, nil
+}
+
+func openPostgres() (*gorm.DB, error) {
```
celinepelletier: Use the existing function `NewPostgresDB`

backend/cmd/createuser/main.go
```diff
@@ -0,0 +1,159 @@
...
+		utils.GetEnvOrDefault("PGHOST", "localhost"),
+		utils.GetEnvOrDefault("PGUSER", "radiant"),
+		utils.GetEnvOrDefault("PGPASSWORD", "radiant"),
+		utils.GetEnvOrDefault("PGDATABASE", "radiant"),
+		utils.GetEnvOrDefault("PGPORT", "5432"),
+		utils.GetEnvOrDefault("PGSSLMODE", "disable"))
+	return gorm.Open(postgres.Open(dsn), &gorm.Config{})
+}
+
+func openStarrocks() (*gorm.DB, error) {
```
celinepelletier: Use existing function `NewStarrocksDB`

backend/compose/scripts/02_starrocks_views.sql
```diff
@@ -0,0 +1,102 @@
...
+   ============================================================================ */
+
+/* --- auth.pii_grant ------------------------------------------------------ */
+CREATE DATABASE IF NOT EXISTS auth;
+DROP VIEW IF EXISTS auth.pii_grant;
+CREATE VIEW auth.pii_grant AS
+SELECT u.user_id AS login, ur.tenant_code, ur.org_code   /* specific-org grants */
+FROM radiant_jdbc.public.user_role ur
+JOIN radiant_jdbc.public.users u
+  ON u.email = ur.email
```
celinepelletier: u.user_id = ur.user_id

backend/compose/scripts/02_starrocks_views.sql
```diff
@@ -0,0 +1,102 @@
...
+JOIN radiant_jdbc.public.role_action ra
+  ON ra.tenant_code = ur.tenant_code AND ra.role_code = ur.role_code
+WHERE ra.action_code = 'can_read_pii' AND ur.org_code <> '*'
+UNION
+SELECT u.user_id AS login, o.tenant_code, o.code         /* '*' expanded to all orgs in tenant */
+FROM radiant_jdbc.public.user_role ur
+JOIN radiant_jdbc.public.users u
+  ON u.email = ur.email
```
celinepelletier: u.user_id = ur.user_id

backend/compose/scripts/02_starrocks_views.sql
```diff
@@ -0,0 +1,102 @@
...
+DROP VIEW IF EXISTS tenant_a.patient;
+CREATE VIEW tenant_a.patient AS
+SELECT id,
+       submitter_patient_id AS mrn,
+       first_name,
+       last_name,
+       date_of_birth,
+       sex_code,
+       organization_code,
+       tenant_code AS row_tenant_code,
```
celinepelletier: why renaming it?

backend/compose/scripts/02_starrocks_views.sql
```diff
@@ -0,0 +1,102 @@
...
+JOIN radiant_jdbc.public.organization o
+  ON o.tenant_code = ur.tenant_code
+WHERE ra.action_code = 'can_read_pii' AND ur.org_code = '*';
+
+/* --- Per-tenant patient views -------------------------------------------- */
+CREATE DATABASE IF NOT EXISTS tenant_a;
+DROP VIEW IF EXISTS tenant_a.patient;
+CREATE VIEW tenant_a.patient AS
+SELECT id,
+       submitter_patient_id AS mrn,
```
celinepelletier: same question

backend/compose/scripts/02_starrocks_views.sql
```diff
@@ -0,0 +1,102 @@
...
+       date_of_birth,
+       sex_code,
+       organization_code,
+       tenant_code AS row_tenant_code,
+       /* convenience flag: does the current login hold can_read_pii for this
+          row's org? Same condition as the mask, so 1 = PII visible (this row is
+          unmasked for me), 0 = masked. NB: reflects the can_read_pii *action*;
+          admins see clear via admin_role bypass and will read 0 here. */
+       EXISTS (SELECT 1 FROM auth.pii_grant g
+               WHERE g.login = substring_index(substr(current_user(), 2), char(39), 1)
```
celinepelletier: was it for the email ?

backend/compose/scripts/02_starrocks_views.sql
```diff
@@ -0,0 +1,102 @@
...
+SELECT id,
+       submitter_patient_id AS mrn,
+       first_name,
+       last_name,
+       date_of_birth,
+       sex_code,
+       organization_code,
+       tenant_code AS row_tenant_code,
+       EXISTS (SELECT 1 FROM auth.pii_grant g
+               WHERE g.login = substring_index(substr(current_user(), 2), char(39), 1)
```
celinepelletier: same question

backend/internal/client/admin_keycloak.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// KeycloakConfigFromEnv builds a KeycloakConfig from the environment, defaulting
+// to the local compose stack. KEYCLOAK_HOST/REALM match the backend's own config;
+// KC_ADMIN_USER/PASS match the admin seeded by master.json.
+func KeycloakConfigFromEnv() KeycloakConfig {
+	return KeycloakConfig{
+		BaseURL:   utils.GetEnvOrDefault("KEYCLOAK_HOST", "http://localhost:8080"),
+		Realm:     utils.GetEnvOrDefault("KEYCLOAK_REALM", "CQDG"),
+		AdminUser: utils.GetEnvOrDefault("KC_ADMIN_USER", "kcadmin"),
```
celinepelletier: Add them also in env.template file (Claude can't access to it, you will have to do it manually)

backend/internal/client/admin_keycloak.go
```diff
@@ -0,0 +1,243 @@
...
+		return "", err
+	}
+
+	user := keycloakUser{
+		Username:        username,
+		Email:           email,
+		FirstName:       firstName,
+		LastName:        lastName,
+		Enabled:         true,
+		EmailVerified:   true,
```
celinepelletier, line 84: Well technically email isn't verified yet...

backend/internal/client/admin_keycloak.go
```diff
@@ -0,0 +1,243 @@
...
+		RequiredActions: []string{},
```
celinepelletier, line 85: Probably add a verify email action

backend/internal/client/admin_keycloak.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	id, err := c.findUserID(ctx, token, username)
+	if err != nil {
+		return "", err
```
celinepelletier, line 90: should the function return (*string, error) to return nil instead of empty string ?

backend/internal/client/admin_keycloak.go
```diff
@@ -0,0 +1,243 @@
...
+		return "", fmt.Errorf("find user %q failed: HTTP %d: %s", username, resp.StatusCode, string(body))
+	}
+	var users []keycloakUser
+	if err := json.Unmarshal(body, &users); err != nil {
+		return "", fmt.Errorf("parse user search for %q: %w", username, err)
+	}
+	if len(users) == 0 {
+		return "", nil
+	}
+	return users[0].ID, nil
```
celinepelletier, line 176: Any chance we have more than 1 result ?

backend/internal/client/admin_keycloak.go
```diff
@@ -0,0 +1,243 @@
...
+		return "", fmt.Errorf("parse admin token: %w", err)
+	}
+	if parsed.AccessToken == "" {
+		return "", fmt.Errorf("admin token response had no access_token")
+	}
+	return parsed.AccessToken, nil
+}
+
+// findUserID returns the id of the user with an exact username match, or "" if none.
+func (c *KeycloakAdminClient) findUserID(ctx context.Context, token, username string) (string, error) {
```
celinepelletier, line 156: Same question ,maybe return *string, error so be able to return nil

backend/internal/repository/auth.go
```diff
@@ -161,3 +161,41 @@ func appendUnique(values []string, value string) []string {
...
 }
+
+// UpsertUser inserts the user or, on email conflict, converges its user_id (the
+// Keycloak sub) and name. This is the write side used at provisioning time, paired
+// with the read side above. Idempotent.
+func (r *AuthRepository) UpsertUser(email, userID, firstName, lastName string) error {
+	err := r.db.Exec(`
+		INSERT INTO public.users (email, user_id, first_name, last_name)
+		VALUES (?, ?, ?, ?)
+		ON CONFLICT (email) DO UPDATE
```
celinepelletier: user_id

backend/internal/repository/auth.go
```diff
@@ -161,3 +161,41 @@ func appendUnique(values []string, value string) []string {
...
+		SET user_id = EXCLUDED.user_id,
```
celinepelletier: no override of the user_id, override email instead

backend/internal/repository/auth.go
```diff
@@ -161,3 +161,41 @@ func appendUnique(values []string, value string) []string {
...
+		    first_name = EXCLUDED.first_name,
+		    last_name = EXCLUDED.last_name`,
+		email, userID, firstName, lastName).Error
+	if err != nil {
+		return fmt.Errorf("upsert user %q: %w", email, err)
```
celinepelletier: user_id, we don't want to log user email that's private information

backend/internal/repository/auth.go
```diff
@@ -161,3 +161,41 @@ func appendUnique(values []string, value string) []string {
...
+	if orgCode != "" {
+		org = orgCode
+	}
+	err := r.db.Exec(`
+		INSERT INTO public.user_role (email, tenant_code, org_code, role_code, granted_by)
+		VALUES (?, ?, ?, ?, 'createuser')
+		ON CONFLICT DO NOTHING`,
+		email, tenantCode, org, roleCode).Error
+	if err != nil {
+		return fmt.Errorf("grant %s/%s/%s to %q: %w", tenantCode, orgCode, roleCode, email, err)
```
celinepelletier: same, no email in logs

backend/internal/repository/auth.go
```diff
@@ -161,3 +161,41 @@ func appendUnique(values []string, value string) []string {
...
+// as NULL (a tenant-wide grant); "*" and specific codes are stored verbatim. The
+// role must already exist (FK to public.role). Idempotent via ON CONFLICT DO NOTHING,
+// which catches either partial-unique index (NULL vs non-NULL org_code).
+func (r *AuthRepository) GrantRole(email, tenantCode, orgCode, roleCode string) error {
+	var org any
+	if orgCode != "" {
+		org = orgCode
+	}
+	err := r.db.Exec(`
+		INSERT INTO public.user_role (email, tenant_code, org_code, role_code, granted_by)
```
celinepelletier: user_id not email

backend/internal/repository/auth.go
```diff
@@ -161,3 +161,41 @@ func appendUnique(values []string, value string) []string {
...
+		VALUES (?, ?, ?, ?, 'createuser')
```
celinepelletier: Probably GrantRole should have another param to fill the granted_by column instead of hardcoded string

backend/internal/repository/auth_test.go
```diff
@@ -322,3 +324,106 @@ func Test_AuthRepository_GetMemberships_MultipleTenantsNoCollision(t *testing.T)
...
+			"INSERT INTO public.role_action (tenant_code, role_code, action_code) VALUES ('radiant', ?, ?) ON CONFLICT DO NOTHING",
+			roleCode, action).Error)
+	}
+}
+
+// purge removes everything a test created, in FK order. Deleting the role cascades
+// its role_action rows. users/user_role are not covered by testutils.cleanUp, so
+// the test owns this cleanup.
+func purge(db *gorm.DB, email, roleCode string) {
+	db.Exec("DELETE FROM public.user_role WHERE email = ?", email)
```
celinepelletier: change for user_id

backend/internal/repository/auth_test.go
```diff
@@ -322,3 +324,106 @@ func Test_AuthRepository_GetMemberships_MultipleTenantsNoCollision(t *testing.T)
...
+	db.Exec("DELETE FROM public.users WHERE email = ?", email)
+	db.Exec("DELETE FROM public.role WHERE tenant_code = 'radiant' AND code = ?", roleCode)
+}
+
+func Test_AuthRepository_UpsertUser_InsertsThenConvergesUserID(t *testing.T) {
+	testutils.RunTest(t, testutils.Need{Postgres: testutils.WritePostgres}, func(t *testing.T, env *testutils.Env) {
+		const email = "createuser-upsert@provisioning.test"
```
celinepelletier: user_id

backend/internal/repository/auth_test.go
```diff
@@ -322,3 +324,106 @@ func Test_AuthRepository_GetMemberships_MultipleTenantsNoCollision(t *testing.T)
...
+		defer purge(env.Postgres, email, "")
+		repo := NewAuthRepository(env.Postgres)
+
+		require.NoError(t, repo.UpsertUser(email, "sub-1", "First", "Last"))
```
celinepelletier: Not sure this test is still accurate since no more unique constraint on email

backend/internal/repository/auth_test.go
```diff
@@ -322,3 +324,106 @@ func Test_AuthRepository_GetMemberships_MultipleTenantsNoCollision(t *testing.T)
...
+
+func Test_AuthRepository_GrantRole_IsIdempotent(t *testing.T) {
+	testutils.RunTest(t, testutils.Need{Postgres: testutils.WritePostgres}, func(t *testing.T, env *testutils.Env) {
+		const email, role = "createuser-grant-idem@provisioning.test", "cu_role_idem"
+		defer purge(env.Postgres, email, role)
+		seedRole(t, env.Postgres, role)
+		repo := NewAuthRepository(env.Postgres)
+		require.NoError(t, repo.UpsertUser(email, "sub-idem", "First", "Last"))
+
+		require.NoError(t, repo.GrantRole(email, "radiant", "ORG_X", role))
```
celinepelletier: adapt to use same user_id

backend/internal/repository/auth_test.go
```diff
@@ -322,3 +324,106 @@ func Test_AuthRepository_GetMemberships_MultipleTenantsNoCollision(t *testing.T)
...
+		require.NoError(t, repo.GrantRole(email, "radiant", "ORG_X", role)) // re-run
+
+		var count int64
+		require.NoError(t, env.Postgres.Raw(
+			"SELECT count(*) FROM public.user_role WHERE email = ? AND tenant_code = 'radiant' AND role_code = ? AND org_code = 'ORG_X'",
```
celinepelletier: user_id

backend/internal/repository/auth_test.go
```diff
@@ -322,3 +324,106 @@ func Test_AuthRepository_GetMemberships_MultipleTenantsNoCollision(t *testing.T)
...
+			email, role).Scan(&count).Error)
+		assert.Equal(t, int64(1), count, "re-grant does not duplicate the row")
+	})
+}
+
+func Test_AuthRepository_GrantRole_TenantWideStoresNullOrg(t *testing.T) {
+	testutils.RunTest(t, testutils.Need{Postgres: testutils.WritePostgres}, func(t *testing.T, env *testutils.Env) {
+		const email, role = "createuser-grant-tenantwide@provisioning.test", "cu_role_tw"
```
celinepelletier: same use user_id instead of email

backend/internal/service/admin.go
```diff
@@ -0,0 +1,113 @@
...
+			return sub, fmt.Errorf("postgres: grant %s/%s/%s to %q: %w", g.TenantCode, g.OrgCode, g.RoleCode, in.Email, err)
+		}
+	}
+
+	if err := deps.Ranger.EnsureUser(ctx, sub); err != nil {
+		return sub, fmt.Errorf("ranger: ensure user %q: %w", sub, err)
+	}
+	for _, tenant := range distinctTenants(in.Grants) {
+		role := RangerTenantRole(tenant)
+		if err := deps.Ranger.AddUserToRole(ctx, role, sub); err != nil {
```
celinepelletier, line 78: Probably we should add some validation like if the role includes at least an org scope action then the role should be applied for at least an org but maybe in another pull request

celinepelletier (changes requested):

backend/internal/service/admin.go
jecos: i dont think it should be done here. Maybe in the caller (service)

backend/internal/client/admin_keycloak.go
jecos: I think it is safe to assume if an admin create a user than the email is verified

backend/internal/client/admin_keycloak.go
jecos: I dont think it is useful

backend/internal/client/admin_keycloak.go
jecos: Well actually it's an antipattern in golang :
https://go.dev/wiki/CodeReviewComments#in-band-errors
I can add a boool found in return type but i' pretty clear for this use case than en empty id is not found

backend/compose/scripts/02_starrocks_views.sql
jecos: no, current_user() return :
'alice'@'%'

backend/compose/scripts/02_starrocks_views.sql
jecos: no, current_user() return :
'alice'@'%'

backend/internal/client/admin_keycloak.go
jecos: See my comment below.

backend/internal/client/admin_keycloak.go
jecos: it shouldn't but i will raise an error

.claude/settings.json
```diff
@@ -13,6 +13,7 @@
       "Write(**/.env)",
       "Write(.env.*)",
       "Write(**/.env.*)"
-    ]
+    ],
+    "allow": ["Read((**/.env.template)", "Edit((**/.env.template)"]
```
LysianeBouchard: Small typo here: `Read((` -> `Read(` and `Edit(( -> Edit(`

backend/internal/service/admin.go
```diff
@@ -0,0 +1,118 @@
+// Package provisioning creates a regular (non-admin) user end-to-end across the
```
LysianeBouchard: Possible that it it "Package service" here?

backend/internal/client/admin_keycloak.go
```diff
@@ -0,0 +1,246 @@
...
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/radiant-network/radiant-api/internal/utils"
+)
+
+// KeycloakConfig points the admin client at a realm and the bootstrap admin used
```
LysianeBouchard, line 16: This comment is not consistent with the AdminUser field comment. I think that this is not a bootstrap admin anymore?

backend/cmd/createuser/main.go
```diff
@@ -0,0 +1,132 @@
...
+		Username: name, Email: *email, FirstName: *first, LastName: *last,
+		Password: password, Grants: grants,
+	}
+
+	ctx := context.Background()
+	sub, err := service.ProvisionUser(ctx, deps, in, grantedBy)
+	if err != nil {
+		log.Fatalf("createuser: provision %q: %v", in.Email, err)
+	}
+	log.Printf("provisioned %s (sub=%s) with %d grant(s)", in.Email, sub, len(in.Grants))
```
celinepelletier, line 83: Unmasked email in logs should be avoid, maybe change it for the Username or mask it partially

backend/compose/scripts/02_starrocks_views.sql
celinepelletier: In production it will be the sub right?

backend/internal/client/admin_keycloak.go
celinepelletier: It could be a way to verify that the user has accepted the invite but I guess we can fine tune that when implementing the user creation feature

backend/internal/service/admin.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return "", fmt.Errorf("keycloak: upsert user %q: %w", in.Username, err)
+	}
+
+	if err := deps.Auth.UpsertUser(sub, in.Email, in.FirstName, in.LastName); err != nil {
+		return sub, fmt.Errorf("postgres: upsert user %q: %w", in.Email, err)
+	}
+	for _, g := range in.Grants {
+		if err := deps.Auth.GrantRole(sub, g.TenantCode, g.OrgCode, g.RoleCode, grantedBy); err != nil {
+			return sub, fmt.Errorf("postgres: grant %s/%s/%s to %q: %w", g.TenantCode, g.OrgCode, g.RoleCode, in.Email, err)
```
celinepelletier: Emails in logs should be masked

backend/internal/service/admin.go
```diff
@@ -0,0 +1,102 @@
...
+// provisioning) — the createuser CLI passes "createuser"; a POST /users handler
+// would pass the acting admin's identity.
+func ProvisionUser(ctx context.Context, deps AdminDeps, in types.UserInput, grantedBy string) (string, error) {
+	sub, err := deps.Keycloak.UpsertUser(ctx, in.Username, in.Email, in.FirstName, in.LastName, in.Password)
+	if err != nil {
+		return "", fmt.Errorf("keycloak: upsert user %q: %w", in.Username, err)
+	}
+
+	if err := deps.Auth.UpsertUser(sub, in.Email, in.FirstName, in.LastName); err != nil {
+		return sub, fmt.Errorf("postgres: upsert user %q: %w", in.Email, err)
```
celinepelletier: Emails in logs should be masked

backend/internal/repository/starrocks_user.go
```diff
@@ -0,0 +1,86 @@
...
+type StarrocksJWTConfig struct {
+	JwksURL          string
+	RequiredIssuer   string
+	RequiredAudience string
+}
+
+// StarrocksJWTConfigFromEnv builds the JWT config from the environment.
+func StarrocksJWTConfigFromEnv() StarrocksJWTConfig {
+	return StarrocksJWTConfig{
+		JwksURL:          utils.GetEnvOrDefault("SR_JWT_JWKS_URL", "http://keycloak:8080/realms/CQDG/protocol/openid-connect/certs"),
```
celinepelletier: Those are missing in env.template

backend/internal/database/starrocks.go
```diff
@@ -20,6 +20,8 @@ var (
 	dbName     = os.Getenv("DB_NAME")
 	dbUserName = os.Getenv("DB_USERNAME")
 	dbPassword = os.Getenv("DB_PASSWORD")
+	dbSSLCA    = os.Getenv("DB_SSL_CA")
```
celinepelletier, line 23: DB_SSL_CA is missing in .env.template

backend/go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/szuecs/gin-glog v1.1.1
 	github.com/tbaehler/gin-keycloak v1.6.1
 	github.com/testcontainers/testcontainers-go v0.41.0
+	golang.org/x/term v0.40.0
```
celinepelletier, line 26: why is it for?

celinepelletier (changes requested):

backend/go.mod
jecos: prompt user password from cli

backend/cmd/createuser/main.go
jecos: Here i think this is fine : it's a command line tool

backend/internal/client/admin_keycloak.go
```diff
@@ -0,0 +1,246 @@
...
+
+// KeycloakConfigFromEnv builds a KeycloakConfig from the environment, defaulting
+// to the local compose stack. KEYCLOAK_HOST/REALM match the backend's own config;
+// KC_ADMIN_USER/PASS match the admin seeded by master.json.
+func KeycloakConfigFromEnv() KeycloakConfig {
+	return KeycloakConfig{
+		BaseURL:   utils.GetEnvOrDefault("KEYCLOAK_HOST", "http://localhost:8080"),
+		Realm:     utils.GetEnvOrDefault("KEYCLOAK_REALM", "CQDG"),
+		AdminUser: utils.GetEnvOrDefault("KC_ADMIN_USER", "kcadmin"),
+		AdminPass: utils.GetEnvOrDefault("KC_ADMIN_PASS", "admin"),
```
LysianeBouchard: [Non blocking]
Claude tells me that it is not a good idea to hardcode a default value for the admin password. The reason is that if the KC_ADMIN_PASS variable is not set, according to Claude, the createuser tool will silently proceeds with a guessable, source-baked admin credential. Claude recommends to fail when the variable is missing.

I thought that it could be a good idea to share, so that we are more aware of this fragility, if Claude turns out to be right of course. Leaving it as optional.

backend/internal/client/admin_keycloak.go
```diff
@@ -0,0 +1,246 @@
...
+	return users[0].ID, nil
+}
+
+func (c *KeycloakAdminClient) createUser(ctx context.Context, token string, user keycloakUser) error {
+	endpoint := fmt.Sprintf("%s/admin/realms/%s/users", c.cfg.BaseURL, c.cfg.Realm)
+	resp, body, err := c.adminRequest(ctx, http.MethodPost, endpoint, token, user)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusConflict {
```
LysianeBouchard, line 185: I am not sure why we tolerate http.StatusConflict here.   I think that createUser is only called after findUserId returned empty. If I am right, the user shouldn't already exist ...

Maybe I missed another use case thought or there is another valid reason ...

backend/internal/client/admin_keycloak.go
```diff
@@ -0,0 +1,246 @@
...
+
+	id, err := c.findUserID(ctx, token, username)
+	if err != nil {
+		return "", err
+	}
+	if id == "" {
+		if err := c.createUser(ctx, token, user); err != nil {
+			return "", err
+		}
+		if id, err = c.findUserID(ctx, token, username); err != nil {
```
LysianeBouchard, line 96: [Non blocking]
Claude recomment to avoid the second findUserID. He thinks that we could check the "Location header" in the createUser response instead. It is an url looking like this: (.../users/{id}).
He might be a bit off... Sorry for the false positive if it is the case.

LysianeBouchard (commented):
I've only done a partial review so far. 

Given the size and scope, I don't think it's the best use of time for me to go deeper right now, so I'm going to pause here unless you'd specifically like me to do a final pass / formal approval.

Left a few comments along the way. All optional, nothing blocking.

backend/internal/client/admin_keycloak.go
jecos: I think it's an hallucination : https://www.keycloak.org/docs-api/latest/rest-api/index.html#_post_adminrealmsrealmusers

I've found nothing in official doc about this.

celinepelletier (approved):

PR #1405, opened 2026-06-12
feat: SJRA-1633 create tenant process

celinepelletier:
## 📌 Context

Adds **tenant provisioning** for the multi-tenancy epic: control-plane tooling that
onboards a tenant across PostgreSQL, StarRocks, and Apache Ranger, plus an automatic
per-tenant StarRocks **view refresh** that keeps views in sync after schema migrations.

Per the ADR (`docs/adr/multi-tenancy-security.md`), each tenant gets its own StarRocks
database (`<code>_tenant`) exposing the clinical data via views over the `radiant_jdbc`
federation, gated by a per-tenant Ranger role/policy. This PR builds that provisioning +
view layer. The API does **not** read these views yet (it still queries
`radiant_jdbc.public.*` directly) — that switch is a separate follow-up, so the startup
refresh stays off by default.

## 📝 Changes

**Two focused control-plane commands** (administrative tools, not the request path):

- **`cmd/create-tenant`** — creates one tenant: `-code -name [-dry-run]`. Runs the
  3-phase, idempotent `service.CreateTenant` (PostgreSQL → StarRocks → Ranger). Needs
  PG + StarRocks + Ranger. `-dry-run` prints the full plan.
- **`cmd/refresh-views`** — re-applies views, one tenant (`-code`) or all. Idempotent;
  needs only PG + StarRocks (no Ranger). The manual / break-glass counterpart to the
  startup refresh.

## ☑️ TO-DOs

- [ ] None blocking merge

## 🧪 Tests

- Test suite full green
- Manual, against the local compose stack

## 🔗 Related Issues

- [SJRA-1633](https://d3b.atlassian.net/browse/SJRA-1633)

## 👀 Notes for Reviewers

Most important parts, in order:

1. **`cmd/api/view_refresh.go` — startup refresh (trickiest).** Runs on boot **before
   serving, on purpose** (a new build must not serve against views that don't match the
   new schema). Gated (default off) + only when a migration changed the schema;
   serialized across replicas by a **session-level** `pg_try_advisory_lock` (deliberately
   not an xact lock, so the StarRocks DDL runs outside any PG transaction → no
   idle-in-transaction); errors are non-fatal. Check the lock acquire/release (unlock
   under `context.WithoutCancel`) and the intentional blocking trade-off.

2. **`internal/repository/starrocks_tenant.go` — `BuildViewStatements` (the view model).**
   (a) No `SELECT *` — the JDBC catalog can't map `jsonb`/`uuid`/`json`, so columns are
   introspected and filtered. (b) A StarRocks view is **not** a Ranger access boundary
   (StarRocks #72910) — tenant isolation is meant to come from the DB-level access policy
   + per-tenant base data, documented in the code and `cmd/create-tenant/README.md`.

3. **`internal/client/admin_ranger.go`.** `EnsureAccessPolicy` is an idempotent upsert
   (no delete-then-create window that could revoke access on failure). `EnsureRole`
   creates if absent — note it treats Ranger's **400** "doesn't have permissions to get
   details" (its confusing response for a missing role) as absent, alongside 404.

4. **Migration `000012` + `internal/repository/tenant.go DefaultRoles`.**
   `can_download_file` now sits on **geneticist**. Caveat: editing 000012 only affects
   fresh DBs; deployed (test) envs are fixed manually.

5. **`internal/service/tenant.go` — `CreateTenant`.** Phase ordering is load-bearing
   (Ranger policy references the DB from the StarRocks phase; `auth.pii_grant` before the
   patient view). Federatable columns are fetched once and threaded through the refresh.

**Out of scope / known:** the API still reads `radiant_jdbc.public.*` directly — the
`FederationName → <tenant>_tenant.*` swap is a separate ticket, so these views + Ranger
policies are infrastructure ahead of their consumer, and `VIEW_REFRESH_ON_STARTUP_ENABLED`
stays off until then. Ranger policies are inert against views (#72910) until enforcement
moves to base tables.

[SJRA-1633]: https://d3b.atlassian.net/browse/SJRA-1633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/cmd/api/view_refresh.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+const viewRefreshOnStartupEnabledEnv = "VIEW_REFRESH_ON_STARTUP_ENABLED"
+
+// viewRefreshLockKey serializes the startup refresh across replicas via a Postgres
+// advisory lock; any stable int64 works.
+const viewRefreshLockKey int64 = 0x726164696e7401
+
+func shouldRefreshViewsOnStartup(enabled, migrated bool) bool {
+	return enabled && migrated
```
jecos, line 23: I'm not sure we need a migration for re-creating views. For instance : if a view using custom sql (like the one defined in patient.tmpl.sql), we need to refresh views on deployment.

backend/cmd/api/view_refresh.go
```diff
@@ -0,0 +1,84 @@
...
+func maybeRefreshTenantViewsOnStartup(ctx context.Context, pg, sr *gorm.DB, migrated bool) {
+	if !shouldRefreshViewsOnStartup(utils.GetBoolEnvOrDefault(viewRefreshOnStartupEnabledEnv, false), migrated) {
+		return
+	}
+	if pg == nil || sr == nil {
+		glog.Warning("startup tenant view refresh skipped: postgres or starrocks connection unavailable")
+		return
+	}
+
+	lister := repository.NewTenantRepository(pg)
```
jecos: what is a lister?

backend/internal/service/tenant.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"github.com/radiant-network/radiant-api/internal/types"
+)
+
+type TenantStore interface {
+	EnsureTenant(code, name string) error
+	SeedDefaultRoles(tenantCode string) error
+}
+
+type TenantLister interface {
```
jecos: suggestion: rename to TenantStore

jecos (changes requested):

backend/cmd/api/main.go
```diff
@@ -300,6 +300,8 @@ func main() {
 	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
 
+	maybeRefreshTenantViewsOnStartup(ctx, dbPostgres, dbStarrocks, migrated)
```
kritchie: It could be a good idea to validate that StarRocks is a valid instance before attempting to call the refresh mechanism (it's pointless to call the method if we already know StarRocks is `nil` and will fail the method) and maybe `log.Fatal` if the refresh fails ? (Likely a symptom of a real problem)

backend/internal/repository/starrocks_tenant.go
```diff
@@ -0,0 +1,150 @@
...
+// BuildAuthStatements builds the global auth database + pii_grant view DDL.
+func BuildAuthStatements() []string {
+	return []string{
+		"CREATE DATABASE IF NOT EXISTS auth",
+		"DROP VIEW IF EXISTS auth.pii_grant",
+		readView("auth_pii_grant.sql"),
+	}
+}
+
+// BuildViewStatements builds the idempotent (DROP+CREATE) DDL for a tenant's database
```
kritchie: Any way we could use the `CREATE [OR REPLACE] VIEW [IF NOT EXISTS]` instead of dropping before creating ? Maybe there are some use cases I'm missing 🤔

backend/cmd/api/view_refresh.go
celinepelletier: We could still use the refresh-view command 🤔 

Reload all the views takes some time so I wouldn't do it at every deployment by default

backend/cmd/api/view_refresh.go
celinepelletier: List all the tenants

backend/internal/service/tenant.go
celinepelletier: They're used by two different functions with zero method overlap:

- CreateTenant writes the tenant row + seeds roles. It never lists tenants.
- RefreshAllTenantViews lists tenants to fan out. It never writes the row or seeds roles.

If you fold ListTenants into TenantStore, then RefreshAllTenantViews — which only needs to list — would have to accept a TenantStore carrying EnsureTenant/SeedDefaultRoles it never calls. That's exactly the interface-segregation smell: a consumer depending on methods it doesn't use. ("The bigger the interface, the weaker the abstraction.")

backend/cmd/api/main.go
celinepelletier: I think Jeremy covers this here: https://github.com/radiant-network/radiant-portal/pull/1410/changes#diff-6c87ae9b5f8c4c8a237c82998a78047b387ecf1d5a4ab2101d8ce3da0600f986R279

backend/internal/repository/starrocks_tenant.go
celinepelletier: Done ✅

backend/internal/service/tenant.go
jecos: My bad I mean TenantReader, store is for persist, reader for read only

backend/cmd/api/view_refresh.go
```diff
@@ -0,0 +1,84 @@
...
+			glog.Errorf("startup tenant view refresh: failed to release advisory lock: %v", err)
+		}
+	}()
+
+	codes, err := service.RefreshAllTenantViews(ctx, lister, lister, srProvisioner)
+	if err != nil {
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			glog.Warningf("startup tenant view refresh aborted (shutting down): %v", err)
+		} else {
+			glog.Errorf("startup tenant view refresh failed (non-fatal): %v", err)
```
kritchie: Since we now fail-fast at startup when StarRocks is not reachable, maybe the error level here could be moved to `fatal` as well ?

kritchie (approved):
FYI focused the review on what I think were the most critical parts. 
I did high level review of the rest. 
This LGTM ! 🚀 ✅

backend/cmd/api/view_refresh.go
jecos: actually slog.Error + osExit(1)

celinepelletier: Fixed ✅

backend/internal/service/tenant.go
celinepelletier: Fixed ✅

jecos (approved):

PR #1410, opened 2026-06-16
feat: SJRA-1590 structured logging & request tracing + SJRA-1594 StarRocks connection handling

jecos:
# feat: SJRA-1590 structured logging & request tracing + SJRA-1594 StarRocks connection handling

## 📌 Context

This branch bundles two related backend reliability/observability improvements:

- **SJRA-1590** — Move the API/worker onto structured, traceable logging so requests can be correlated end-to-end in JSON logs.
- **SJRA-1594** — Make the API recover predictably when StarRocks is unreachable, instead of booting into an unrecoverable state that panics on every request.

## 📝 Changes

**SJRA-1590 — structured logging & traceable request id**
- Switch logging to stdlib `log/slog` with JSON output, centralized in a new `internal/observability` package (`logging.go` + tests). `LOG_LEVEL` env var (`debug`/`info`/`warn`/`error`, default `info`).
- New `RequestID` middleware: reuses an inbound `X-Request-ID` or mints a UUID, echoes it on the response, and threads it through the request context so every `slog.*Context` line carries `request_id`.
- `HandleError` returns the same correlation id as `X-Correlation-ID`.
- Middleware stack reordered: request id → structured request logging → metrics → gzip → Keycloak logger → CORS → role-based auth → recovery. Prometheus metrics exposed at public `GET /metrics`.
- Repository/DB logging migrated from `log.Print`/`log.Fatal` to `slog` (incl. `CheckDatabaseConnection` now `slog.Error` + `return "down"` rather than `log.Fatal`, so a runtime DB blip no longer kills the process).

**SJRA-1594 — better StarRocks connection handling**
- `cmd/api/main.go`: `NewStarrocksDB()` failure now `os.Exit(1)` (mirrors the existing PostgreSQL handling). The API never boots with a nil DB; the orchestrator restarts it cleanly once StarRocks is back.
- Removed the now-dead `if db == nil { return nil }` guards from all 26 repository constructors (and the `panic` variant in `NewValueSetsRepository` + the `pubmedClient == nil` guard in `NewInterpretationsRepository`). With the DB guaranteed non-nil at wiring time, these branches were unreachable and were the source of a confusing deferred `nil pointer dereference` panic far from the real cause (a dead DB). Constructors now return the concrete struct unconditionally, matching `NewPostgresRepository`. Unused `log/slog` imports dropped.

## ☑️ TO-DOs

- [ ] None

## 🧪 Tests

- `go build ./...`, `make fmt`, `make lint` — all clean (0 lint issues).
- New tests: `internal/observability/logging_test.go`, `internal/server/middlewares_request_test.go`.
- Manual end-to-end verification of SJRA-1594 against the docker-compose stack:
  - StarRocks unreachable at startup → API process exits with code 1, never serves a request.
  - StarRocks up → `GET /status` → `{"postgres":"up","starrocks":"up"}`.
  - Runtime StarRocks stop → `/status` reports `starrocks:"down"`, API stays alive (no restart); StarRocks restart → `/status` self-heals to `"up"` with the same process (`RestartCount` unchanged).

## 🔌 Behavior when StarRocks is unavailable

The fix cleanly separates two cases:

**StarRocks down at startup** — the API process **exits immediately** (`os.Exit(1)`) and never starts listening. No handler runs; under an orchestrator the pod stays not-Ready / crashloops and is kept out of the load balancer until StarRocks is reachable, then boots cleanly. This is the scenario that previously booted with nil repos and panicked on every request.

**StarRocks drops at runtime** (was up at boot) — the process stays alive (the health check no longer `log.Fatal`s). A call to a StarRocks-backed endpoint:
- does **not** panic — the repository's `*gorm.DB` is still a valid, non-nil pointer; only the underlying connection is dead;
- fails at query execution, the error propagates to the handler, and `HandleError` returns a generic **HTTP 500** `{"status":500,"message":"Internal Server Error"}` with an `X-Correlation-ID` header (the full error is logged server-side via `slog`, never leaked to the client);
- leaves the process running — once StarRocks returns, the gorm connection pool reconnects and subsequent calls succeed **with no restart**.

| StarRocks down… | Result | Recovery |
|---|---|---|
| at startup | process exits (code 1), never serves | orchestrator restarts; clean boot once StarRocks is back |
| at runtime | endpoint returns HTTP 500 + `X-Correlation-ID`, process stays up | self-heals on next query when StarRocks returns |

> Caveat: if StarRocks is black-holed (host unreachable) rather than actively refusing connections, the driver's dial can block until a TCP timeout, so those 500s may return slowly under a hard outage — there is no explicit per-query connect timeout in the DSN (`NewStarrocksDB`). Not a correctness issue; a `timeout=…&readTimeout=…` on the DSN could bound it as a follow-up if desired.

## 🔗 Related Issues

- [SJRA-1590](https://d3b.atlassian.net/browse/SJRA-1590)
- [SJRA-1594](https://d3b.atlassian.net/browse/SJRA-1594)

## 👀 Notes for Reviewers

- The `CheckDatabaseConnection` `log.Fatal` → `slog.Error`/`return "down"` change (under SJRA-1590) and the startup `os.Exit(1)` (SJRA-1594) are complementary: the former lets a *runtime* outage self-heal via the gorm pool; the latter makes a *startup* outage fail fast instead of running degraded.

[SJRA-1590]: https://d3b.atlassian.net/browse/SJRA-1590?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/compose/starrocks-conf/fe.conf
```diff
@@ -100,13 +100,3 @@ ssl_keystore_location = /data/deploy/certs/starrocks_keystore.jks
...
-priority_networks = 127.0.0.1/32
-priority_networks = 127.0.0.1/32
-priority_networks = 127.0.0.1/32
-priority_networks = 127.0.0.1/32
-priority_networks = 127.0.0.1/32
-priority_networks = 127.0.0.1/32
-priority_networks = 127.0.0.1/32
-priority_networks = 127.0.0.1/32
-priority_networks = 127.0.0.1/32
-priority_networks = 127.0.0.1/32
```
jecos, line 112: this was added by compose, we should use ro tag to avoid this

backend/internal/observability/logging.go
jecos, line 1: Here one of the most important file to review

backend/internal/repository/auth.go
```diff
@@ -128,7 +124,11 @@ func applyGrant(membership *types.TenantMembership, grant membershipGrant, tenan
 	case types.ActionScopeOrg:
 		if grant.OrgCode == nil {
 			// Org-scoped action on a tenant-wide grant (org_code NULL) so this action applies to no specific org and is intentionally dropped.
-			log.Printf("WARN: Org-scoped action %s should not be applied tenant wide (org_code = NULL) - Verify the role %s for tenant %s.", grant.ActionCode, grant.RoleCode, grant.TenantCode)
+			slog.Warn("org-scoped action should not be applied tenant wide (org_code = NULL)",
+				slog.String("action", grant.ActionCode),
+				slog.String("role", grant.RoleCode),
+				slog.String("tenant", grant.TenantCode),
+			)
```
jecos, line 131: Using structured logging

backend/internal/repository/auth.go
```diff
@@ -15,10 +15,6 @@ type AuthRepository struct {
 }
 
 func NewAuthRepository(db *gorm.DB) *AuthRepository {
-	if db == nil {
-		log.Print("AuthRepository: db is nil")
-		return nil
-	}
```
jecos, line 21: do not return nil otherwise we will have a panic error on nil pointer reference. instead let repo handled error and return 500

backend/go.mod
```diff
@@ -70,6 +68,7 @@ require (
 	github.com/go-playground/universal-translator v0.18.1 // indirect
 	github.com/go-playground/validator/v10 v10.20.0 // indirect
 	github.com/goccy/go-json v0.10.5 // indirect
+	github.com/golang/glog v1.2.5 // indirect
```
jecos, line 71: glog become transitive (gin-keycloak)

backend/cmd/worker/main.go
```diff
@@ -104,17 +107,18 @@ func processBatch(ctx context.Context, db *gorm.DB, bv *batchval.BatchValidation
...
 		}
 	}
 }
 
 func StartHealthProbe(ctx context.Context, wg *sync.WaitGroup, db *gorm.DB) {
 	port := utils.GetEnvOrDefault("PROBE_PORT", "9999")
 	if _, err := strconv.Atoi(port); err != nil {
-		glog.Fatalf("Probe port defined in env var PROBE_PORT (%v) must be an integer ", port)
+		slog.Error("PROBE_PORT must be an integer", slog.String("value", port))
+		os.Exit(1)
```
jecos, line 121: Fail fast if starrocks is not reachable

backend/cmd/api/main.go
```diff
@@ -276,20 +268,21 @@ func newEngine() *gin.Engine {
...
+	observability.Setup()
 
 	database.MigrateWithEnvDefault()
 
 	// Initialize database connection
 	dbStarrocks, err := database.NewStarrocksDB()
 	if err != nil {
-		log.Print("Failed to initialize starrocks database: ", err)
+		slog.Error("failed to initialize starrocks database", slog.Any("error", err))
+		os.Exit(1)
```
jecos, line 279: Fail fast if Starrocks is not reachable

celinepelletier (approved):

PR #1415, opened 2026-06-17
feat: CLIN-6059 - tenant 403 not found instead of 500

creativeyann17:
- [x] `404` if tenant doesn't exist
- [x] add `QLIN` in init database SQL script
- [x] Unit Tests
- [x] Local Tests

### DIFF
--- a/backend/internal/server/middlewares.go
+++ b/backend/internal/server/middlewares.go
@@ -99,13 +99,15 @@ type tenantAccessChecker interface {
 }
 
 // RequireTenantAccess gates tenant-scoped routes (`/:tenant/...`). It always reads the
-// `tenant` path param, rejects an unknown tenant with 404, and stores the resolved tenant
+// `tenant` path param, rejects an unknown tenant with 403, and stores the resolved tenant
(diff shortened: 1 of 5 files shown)


backend/internal/server/middlewares.go
```diff
@@ -109,6 +114,18 @@ func RequireTenantAccess(auth utils.Auth, repo tenantAccessChecker, enforce bool
...
 		tenant := c.Param("tenant")
 
+		exists, err := repo.TenantExists(tenant)
+		if err != nil {
+			HandleError(c, err)
+			c.Abort()
+			return
+		}
+		if !exists {
+			HandleNotFoundError(c, "tenant")
```
celinepelletier: I wonder if we should just answer a 401 ? Kinda worried to give information to a potential hacker with this 404

creativeyann17: If we want to avoid leaking that info (fair), then it's 403 (same as HasTenantAccess) cause 401 is token related not about access rights

celinepelletier (approved):

PR #1423, opened 2026-06-18
fix: SJRA-1657 retrieve user on email instead of username

celinepelletier:
## 📌 Context

`cmd/createuser` failed when re-run for a user that already exists in Keycloak with a username that differs from their email (e.g. `cypress@email.me` stored with username `cypress`):

```
createuser: provision "[cypress@email.me](mailto:cypress@email.me)": keycloak: upsert user "[cypress@email.me](mailto:cypress@email.me)": user "[cypress@email.me](mailto:cypress@email.me)" not found after create
```

`UpsertUser` looked the user up by **username** (`?username=…&exact=true`), assuming the realm enforces email-as-username. When that assumption doesn't hold, the lookup misses, the create POST conflicts on the unique email (silently accepted as a 409), and the follow-up lookup misses again → `not found after create`. The result: an existing user could never be matched, so granting an existing user access to a new tenant was impossible.

## 📝 Changes

- **Look up users by email instead of username** in `KeycloakAdminClient.UpsertUser` (`findUserID` → `findUserIDByEmail`, querying `?email=…&exact=true`). 
- New users are still **created** with `username = email` via the admin API, which is unaffected by the realm's `registrationEmailAsUsername` setting.

## ☑️ TO-DOs

- [ ] None

## 🧪 Tests

- Updated the test fake to key users by email (`usersByEmail`) and match the `email` query param.
- The existing-user test now uses a deliberate `username != email` case (`username "alice"` / `email "alice@demo.org"`), which reproduces the original failure scenario and verifies the user is matched and updated rather than re-created.
- Run: `cd backend && go test ./internal/client/...`
- Test locally with success to add existing user to a new tenant

## 🔗 Related Issues

- [SJRA-1657](https://d3b.atlassian.net/browse/SJRA-1657)

## 👀 Notes for Reviewers

- The fix is config-agnostic: it works whether the realm is email-as-username or not, because the username search assumption was the bug.
- The one remaining assumption is that **emails are unique** (`duplicateEmailsAllowed: false`, the Keycloak default). If emails are not unique, `findUserIDByEmail` fails loudly with `expected at most 1 exact match` rather than silently picking the wrong user — intended behavior.

[SJRA-1657]: https://d3b.atlassian.net/browse/SJRA-1657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/internal/client/admin_keycloak.go
+++ b/backend/internal/client/admin_keycloak.go
@@ -87,7 +87,10 @@ func (c *KeycloakAdminClient) UpsertUser(ctx context.Context, username, email, f
 		RequiredActions: []string{},
 	}
 
-	id, err := c.findUserID(ctx, token, username)
+	// Match on email, not username: a pre-existing user may have been provisioned
+	// with a username that differs from its email (the realm need not enforce
+	// email-as-username), so email is the reliable identity key.
+	id, err := c.findUserIDByEmail(ctx, token, email)
 	if err != nil {
 		return "", err
 	}
@@ -95,11 +98,11 @@ func (c *KeycloakAdminClient) UpsertUser(ctx context.Context, username, email, f
 		if err := c.createUser(ctx, token, user); err != nil {
 			return "", err
 		}
-		if id, err = c.findUserID(ctx, token, username); err != nil {
+		if id, err = c.findUserIDByEmail(ctx, token, email); err != nil {
 			return "", err
 		}
 		if id == "" {
-			return "", fmt.Errorf("user %q not found after create", username)
+			return "", fmt.Errorf("user %q not found after create", email)
 		}
 	} else {
 		if err := c.updateUser(ctx, token, id, user); err != nil {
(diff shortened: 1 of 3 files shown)


backend/internal/client/admin_keycloak.go
```diff
@@ -87,19 +87,22 @@ func (c *KeycloakAdminClient) UpsertUser(ctx context.Context, username, email, f
...
+	id, err := c.findUserIDByEmail(ctx, token, email)
 	if err != nil {
 		return "", err
 	}
 	if id == "" {
 		if err := c.createUser(ctx, token, user); err != nil {
 			return "", err
 		}
-		if id, err = c.findUserID(ctx, token, username); err != nil {
+		if id, err = c.findUserIDByEmail(ctx, token, email); err != nil {
```
kritchie, line 101: Does it have to be one or the other ? 
Like could we try the email first, and then fallback on the username? 🤔 

**Edit**: Nevermind, I think I misunderstood the problem, all good 👍

kritchie (approved):

backend/internal/client/admin_keycloak.go
celinepelletier: We could but we will never provide the username which is not an email since the command arg is `-email`

PR #1434, opened 2026-06-25
feat: SJRA-1461 — API reads from per-tenant StarRocks views instead of radiant_jdbc

celinepelletier:
# feat: SJRA-1461 — API reads from per-tenant StarRocks views instead of `radiant_jdbc`

## 📌 Context

The API currently reads clinical/operational data straight from the `radiant_jdbc` JDBC catalog (`radiant_jdbc.public.*`). In the target multi-tenancy model each tenant gets its own StarRocks database `<code>_tenant` holding views over the federation (filtered by `tenant_code` where applicable), and the API must address **those views** — never `radiant_jdbc` directly — so that per-user JWT access control can eventually be enforced at the StarRocks/Ranger boundary.

This PR makes the API read path tenant-aware. It is **backward-compatible and gated**: behavior is byte-identical to today until the new flag is turned on.

## 📝 Changes

- **Resolver + table helper** (`internal/types/tenant_context.go`, `table_field.go`): `TenantSchema(ctx)` returns the tenant's `<code>_tenant` database when a tenant is bound to the request context, else the legacy `radiant_jdbc.public`. `Table.In(schema)` qualifies a table by schema (`CaseTable.In("radiant_jdbc.public") == CaseTable.FederationName`).
- **Read-path conversion** of every consumer repository — `cases`, `variants`, `documents`, `sequencing_experiment`, `igv`, `germline_snv`/`somatic_snv`/`germline_cnv` occurrences — plus the shared join helpers in `internal/utils/repositories.go`. Every `FederationName` / raw `radiant_jdbc.public.*` reference now resolves through `TenantSchema(ctx)`. Native StarRocks tables (genomic occurrences/variants, `mondo_term`, staging) are untouched.
- **Gated middleware switch** (`internal/server/middlewares.go`): `RequireTenantAccess` binds the resolved tenant to the request context **only when `TENANT_VIEWS_READ_ENABLED=true`** (new env, default **off**). Off → no tenant in context → reads stay on `radiant_jdbc.public` (unchanged).
- **Full view coverage** (`internal/repository/starrocks_tenant.go`): the provisioner now creates a `<code>_tenant` view for **all** federated tables — tenant-scoped ones filtered by `tenant_code`, reference/value-set/junction tables exposed unfiltered — so the API never needs `radiant_jdbc`.
- **Seed data** (`scripts/init-sql/insert_clinical_data.sql`): add the `qlin` tenant with its `member` / `geneticist` / `data_manager` roles + `role_action` grants (local compose seeds).

## ☑️ TO-DOs

- [ ] After each tenant's views are provisioned & populated (`cmd/create-tenant` / `cmd/refresh-views`), flip `TENANT_VIEWS_READ_ENABLED=true` in the target environments to activate tenant-scoped reads.

## 🧪 Tests

- **Unit**: `TenantSchema`/`TenantFromContext` resolution (including empty-code → federation default), `Table.In` ↔ `FederationName` equivalence, and `RequireTenantAccess` binding on/off (`radiant_tenant` vs `radiant_jdbc.public`). Provisioner view-builder tests updated for the expanded table set.
- **Integration**: the conversion is byte-identical under the default (no tenant in context), so the existing repository suites exercise it unchanged; the variants suite was verified producing identical legacy SQL. Run with `make itest`.
- **Local**: full frontend + backend stack manually verified end-to-end with the flag on
- `go build ./...`, `go vet`, `gofmt`, and `golangci-lint` all clean.

## 🔗 Related Issues

- [SJRA-1461](https://d3b.atlassian.net/browse/SJRA-1461)

## 👀 Notes for Reviewers

- **The defaulting resolver is the safety mechanism**: with `TENANT_VIEWS_READ_ENABLED` off, `TenantSchema(ctx)` returns `radiant_jdbc.public` everywhere, so the generated SQL is unchanged. Worth confirming you're comfortable with that as the rollout gate.
- The shared `Join*` helpers resolve the schema from the context carried on the GORM statement (`tx.Statement.Context`) — that's why the 4 non-`cases` repos needed no call-site changes; they self-migrate when a tenant is present.
- Bare `.Name` table references (writes/lookups like `UpdateCaseDiagnosisLabCode`, `GetCaseBySubmitterCaseIdAndProjectId`) are **not** federation reads and were intentionally left on the default catalog.
- Known StarRocks+Ranger caveat (tracked in the epic): a view is not a Ranger access boundary — isolation will come from the per-tenant DB access policy; this PR is the addressing change that precedes full authz.

[SJRA-1461]: https://d3b.atlassian.net/browse/SJRA-1461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/internal/repository/cases.go
```diff
@@ -113,31 +114,32 @@ func (r *CasesRepository) SearchCases(ctx context.Context, userQuery types.ListQ
 
 func (r *CasesRepository) SearchById(ctx context.Context, prefix string, limit int) (*[]AutocompleteResult, error) {
 	/**
-	  	(SELECT "case_id" as type, id as value from `radiant_jdbc`.`public`.`cases` WHERE CAST(id AS TEXT) LIKE '1%')
+	  	(SELECT "case_id" as type, id as value from cases WHERE CAST(id AS TEXT) LIKE '1%')
 	    UNION
-	    (SELECT "patient_id" as type, proband_id as value from `radiant_jdbc`.`public`.`cases` WHERE CAST(proband_id AS TEXT) LIKE '1%')
+	    (SELECT "patient_id" as type, proband_id as value from cases WHERE CAST(proband_id AS TEXT) LIKE '1%')
```
LysianeBouchard: [MINOR] I think that this is just a comment here and not a real query, but it might be more clear to use `<schema>.case`  instead of just `case`.

backend/internal/repository/germline_snv_occurrences.go
```diff
@@ -48,17 +49,17 @@ func (r *GermlineSNVOccurrencesRepository) GetOccurrences(ctx context.Context, c
...
 	// FROM (germline__snv__occurrence o, snv__variant v)
-	// 		LEFT JOIN (SELECT DISTINCT locus_id, case_id, sequencing_id FROM radiant_jdbc.public.interpretation_germline) i ON i.locus_id = g_snv_o.locus_id AND i.sequencing_id = ?
-	// 		LEFT JOIN (SELECT DISTINCT occurrence_id, case_id, seq_id, task_id FROM radiant_jdbc.public.occurrence_note WHERE deleted = false) note ON note.occurrence_id = g_snv_o.locus_id AND note.task_id = g_snv_o.task_id AND note.seq_id = ? AND note.case_id = ?
-	// 		LEFT JOIN radiant_jdbc.public.occurrence_flag flag ON flag.occurrence_id = g_snv_o.locus_id AND flag.task_id = g_snv_o.task_id AND flag.seq_id = ? AND flag.case_id = ?
+	// 		LEFT JOIN (SELECT DISTINCT locus_id, case_id, sequencing_id FROM interpretation_germline) i ON i.locus_id = g_snv_o.locus_id AND i.sequencing_id = ?
+	// 		LEFT JOIN (SELECT DISTINCT occurrence_id, case_id, seq_id, task_id FROM occurrence_note WHERE deleted = false) note ON note.occurrence_id = g_snv_o.locus_id AND note.task_id = g_snv_o.task_id AND note.seq_id = ? AND note.case_id = ?
```
LysianeBouchard: [Minor]. Here too you might want to use `<schema>.occurence_note`.

LysianeBouchard (approved):
Minor optional comments only.

backend/cmd/api/view_refresh.go
```diff
@@ -19,21 +19,21 @@ const viewRefreshOnStartupEnabledEnv = "VIEW_REFRESH_ON_STARTUP_ENABLED"
 // advisory lock; any stable int64 works.
 const viewRefreshLockKey int64 = 0x726164696e7401
 
-func shouldRefreshViewsOnStartup(enabled, migrated bool) bool {
-	return enabled && migrated
+func shouldRefreshViewsOnStartup(enabled bool) bool {
```
kritchie: This could be removed, I don't think we need this anymore

celinepelletier: it's still conditioned by the toggle feature `VIEW_REFRESH_ON_STARTUP_ENABLED`

backend/cmd/api/view_refresh.go
```diff
@@ -19,21 +19,21 @@ const viewRefreshOnStartupEnabledEnv = "VIEW_REFRESH_ON_STARTUP_ENABLED"
...
+// self-heals drift; gated (default off) and advisory-locked so one replica runs it per deploy.
 //
 // A returned error is fatal to the caller (don't serve a build whose views don't match
-// the schema). Cases that are NOT errors and return nil: the refresh is disabled/not
-// needed, another replica holds the lock, or a shutdown cancelled it mid-run.
-func maybeRefreshTenantViewsOnStartup(ctx context.Context, pg, sr *gorm.DB, migrated bool) error {
-	if !shouldRefreshViewsOnStartup(utils.GetBoolEnvOrDefault(viewRefreshOnStartupEnabledEnv, false), migrated) {
+// the schema). Cases that are NOT errors and return nil: the refresh is disabled, another
+// replica holds the lock, or a shutdown cancelled it mid-run.
+func maybeRefreshTenantViewsOnStartup(ctx context.Context, pg, sr *gorm.DB) error {
```
kritchie: [nit] Maybe move the `GetBoolEnvOrDefault` at the top level and rename this function `refreshTenantViews` to make the intent clearer.

backend/cmd/api/view_refresh.go
kritchie: Yes but more in the sense that it's a passthrough method that returns the boolean value directly without any operations on it. I don't think its useful to keep this method (and the related unit tests)

celinepelletier: Removing the function, we will test the toggle feature in main

kritchie (commented):
LGTM 🚀 
This is nice 👌

PR #1455, opened 2026-07-08
feat: SJRA-1468 add mysql proxy

celinepelletier:
# feat: SJRA-1468 add mysql-proxy for JWT auth to StarRocks

## 📌 Context

Ranger enforces per-user PII masking / row-filter / tenant access at the StarRocks layer, but only when a query runs as the authenticated user. Direct clients (Jupyter, BI, Python ≥9.2) speak StarRocks' `authentication_openid_connect_client` (JWT) plugin natively. But **Go's `go-sql-driver/mysql` implements a fixed set of auth plugins with no extension point**, so GORM (the API/worker) can't log in as a JWT user without forking the driver.

This adds a small TCP proxy that bridges the gap during the login handshake:
Go client ──cleartext (mysql_clear_password: JWT as the "password")──▶ proxy
proxy     ──TLS + authentication_openid_connect_client (JWT as auth data)──▶ StarRocks

After login it's a transparent byte pipe. This is the enabler for running the API's own queries
as the end user (wired in a follow-up) so masking needs no reimplementation in the app.

## 📝 Changes

- **`backend/cmd/mysql-proxy/`** — the proxy: MySQL packet framing, the login-handshake translation (strip `CLIENT_SSL` from the client so it stays cleartext; do TLS + the OIDC auth-switch toward StarRocks), and a transparent pipe. Ported from the POC (`docs/adr/ranger-poc/proxy/`) and hardened: verify StarRocks' cert against `STARROCKS_SSL_CA` (skip-verify fallback for dev), structured `slog`, auth-phase deadlines, graceful shutdown on
  SIGINT/SIGTERM, and an HTTP health probe (`GET /status`).
- **`backend/mysql-proxy.Dockerfile`** — `alpine` runtime (ships `curl` for the Fargate in-container health check), runs as `nobody`, mirrors `api.Dockerfile`'s build.
- **`.github/workflows/build_and_push.yml`** — builds/pushes `radiant-mysql-proxy` (metadata + build-push + provenance attestation) alongside the api/worker images.
- **`backend/.gitignore`** — ignore the root `mysql-proxy` build artifact (anchored so it doesn't ignore the `cmd/mysql-proxy/` source).

## ☑️ TO-DOs

- [ ] Deploy as a **sidecar** in the two QA envs (QLIN/CHOP) — loopback-only, with `STARROCKS_SSL_CA` set to verify StarRocks' cert.
- [ ] Wire the API read path to open per-user proxy connections (separate PR).

## 🧪 Tests

- **Unit** — `proxy_test.go` covers the fiddly byte-math helpers: username extraction, capability stripping, charset parsing, SSL/handshake/OIDC packet builders (incl. the length-encoded path real JWTs hit), auth-switch and error packets.
- **Validated end-to-end against the QLIN Starrocks** — a `go-sql-driver` client (and the API) authenticating through the proxy with a real Keycloak JWT: a member user sees PII masked (`***` / year-only DOB), a geneticist sees it clear; the proxy logs `authenticated, piping`.
- `go build ./...`, `go test ./cmd/mysql-proxy`, `make fmt`, `go vet` all clean.

## 🔗 Related Issues

- [SJRA-1468](https://d3b.atlassian.net/browse/SJRA-1468)

## 👀 Notes for Reviewers

- **Security — cleartext JWT on the client↔proxy hop.** The proxy asks the Go client for `mysql_clear_password`, so the JWT is cleartext to the proxy. It must run **co-located with the API** (same pod/task) and its port **must never be exposed** beyond loopback. Documented in the package doc + README.
- **Why hand-rolled protocol code?** `go-sql-driver` has no pluggable auth (would require a driver fork — more maintenance than this ~370-LOC proxy); StarRocks JWT users have no documented cleartext-direct path. The proxy only parses the *login* handshake (stable); after auth it's a dumb pipe. The tricky byte offsets are commented and unit-tested.
- **Backend TLS**: verified against `STARROCKS_SSL_CA` when set; unset falls back to skip-verify with a WARN (dev only — QA/prod should set it). The expected server name is the host from `STARROCKS_ADDR`.
- **Additive / low-risk to merge**: nothing consumes the proxy yet — the API only uses it where `STARROCKS_PROXY_ADDR` is set, which no shipped path does. Deploying the sidecar and flipping that env is a separate, reversible step.

[SJRA-1468]: https://d3b.atlassian.net/browse/SJRA-1468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/cmd/mysql-proxy/main.go
+++ b/backend/cmd/mysql-proxy/main.go
@@ -1,0 +1,128 @@
+// Command mysql-proxy sits between go-sql-driver clients (the Radiant API/worker via GORM)
+// and StarRocks. Go's MySQL driver can't speak StarRocks' authentication_openid_connect_client
+// plugin, so it can't log in as a JWT user directly. This proxy translates: the Go client
+// connects in cleartext and sends its Keycloak JWT as the "password" (mysql_clear_password),
+// and the proxy performs the real TLS + OIDC handshake to StarRocks on its behalf. Every query
+// then runs as the authenticated user, so Ranger enforces per-user masking / row-filter /
+// tenant access — no need to reimplement any of that in the API.
+//
+// SECURITY: the JWT crosses the client↔proxy hop in cleartext, so run the proxy on the same
+// host (or a trusted network) as the Go service and restrict its listen port to that peer.
+//
+//	LISTEN_ADDR      address to accept cleartext clients on   (default :9031)
+//	STARROCKS_ADDR   StarRocks FE MySQL address               (default starrocks:9030)
+//	STARROCKS_SSL_CA path to a CA bundle to verify StarRocks' cert; unset = skip verification (dev)
+//	PROBE_PORT       HTTP health-probe port                   (default 9998)
+package main
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net"
+	"net/http"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+)
+
+const authTimeout = 15 * time.Second
+
+func main() {
+	slog.SetDefault(slog.New(slog.NewJSONHandler(os.Stdout, nil)))
+
+	listenAddr := getenv("LISTEN_ADDR", ":9031")
+	backendAddr := getenv("STARROCKS_ADDR", "starrocks:9030")
+	probePort := getenv("PROBE_PORT", "9998")
+
+	tlsCfg, err := backendTLSConfig(os.Getenv("STARROCKS_SSL_CA"), backendAddr)
+	if err != nil {
+		slog.Error("build backend TLS config", "error", err)
+		os.Exit(1)
+	}
+	p := &proxy{backendAddr: backendAddr, tlsCfg: tlsCfg, authTimeout: authTimeout}
+
+	// Cancel the root context on SIGINT/SIGTERM so the accept loop and health server stop.
+	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer stop()
+
+	go serveHealthProbe(ctx, probePort)
+
+	ln, err := net.Listen("tcp", listenAddr)
+	if err != nil {
+		slog.Error("listen failed", "addr", listenAddr, "error", err)
+		os.Exit(1)
+	}
+	slog.Info("mysql-proxy listening", "listen", listenAddr, "backend", backendAddr)
+
+	// Unblock Accept on shutdown by closing the listener.
+	go func() {
+		<-ctx.Done()
+		_ = ln.Close()
+	}()
+
+	for {
+		conn, err := ln.Accept()
+		if err != nil {
+			if ctx.Err() != nil {
+				slog.Info("shutting down")
+				return
+			}
+			slog.Error("accept error", "error", err)
+			continue
+		}
+		go p.handle(conn)
+	}
+}
+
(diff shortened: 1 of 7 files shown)


backend/cmd/mysql-proxy/proxy.go
```diff
@@ -0,0 +1,365 @@
...
+	clientProtocol41 uint32 = 0x00000200 // 4.1+ protocol (always on for us)
+	clientSecureConn uint32 = 0x00008000 // secure-auth handshake layout
+	clientPluginAuth uint32 = 0x00080000 // pluggable authentication
+)
+
+const (
+	maxPacket    = 16 * 1024 * 1024                       // max packet size we advertise (standard)
+	nativePlugin = "mysql_native_password"                // classic password plugin
+	clearPlugin  = "mysql_clear_password"                 // send-password-in-clear (what we ask the Go client for)
+	oidcPlugin   = "authentication_openid_connect_client" // StarRocks' JWT plugin (backend side)
```
kritchie: Looks like this is not used anywhere else 🤔, could it be removed ? (Unless maybe it's used implicitly somewhere else)

celinepelletier: Good catch, I removed it :)

backend/cmd/mysql-proxy/main.go
```diff
@@ -0,0 +1,128 @@
...
+		conn, err := ln.Accept()
+		if err != nil {
+			if ctx.Err() != nil {
+				slog.Info("shutting down")
+				return
+			}
+			slog.Error("accept error", "error", err)
+			continue
+		}
+		go p.handle(conn)
```
kritchie: FYI maybe that's what we want, but unless I've missed something if there's a panic in one of the handlers, it crashes the whole proxy. 
(If we don't want this, we could add a recovery in the function)

backend/cmd/mysql-proxy/proxy.go
```diff
@@ -0,0 +1,364 @@
...
+		handshake[high] &^= byte(capability >> 16)
+		handshake[high+1] &^= byte(capability >> 24)
+	}
+	return handshake
+}
+
+// charsetFromHandshake reads the charset byte StarRocks chose (it sits right after the low
+// capability word), so we can echo it back in the packets we build. Defaults to utf8 (0x21).
+func charsetFromHandshake(handshake []byte) byte {
+	charset := serverVersionEnd(handshake) + 4 + 8 + 1 + 2
```
kritchie: [nit] It could be useful to document the magic numbers here.

kritchie (approved):
This is low-level stuff and outside of my main expertise. 
So I'll approve, but maybe you should get someone else's feedback as well. 

Otherwise, LGTM 👍

backend/cmd/mysql-proxy/main.go
celinepelletier: Good catch — fixed. Added a per-connection recover (recoverConn) so a panic in one handler goroutine logs + drops just that connection instead of crashing the process (same pattern as net/http). Also hardened the one real panic vector it would've hit: the ERR-packet branch now guards against a truncated authResult before reading the 2-byte code. Covered by a test.

backend/cmd/mysql-proxy/proxy.go
celinepelletier: 👍🏻  Replaced the offset magic numbers with named hs*Len constants

PR #1458, opened 2026-07-09
feat: CLIN-6022 onset_code / interpretation_code optional for ancestry/consanguinity observations

emmanuelnau:
## 📌 Context

The case-batch contract (`POST /{tenant}/cases/batch`) required `onset_code` and
`interpretation_code` on **every** categorical observation. This blocks observations that
have neither an onset nor a positive/negative interpretation — namely **ancestry**
(ethnicity) and **consanguinity**.

These two fields are now required for **every** categorical observation **except** the
`ancestry` and `consanguinity` codes, which are exempt.

## 📝 Changes

- **Exemption keyed on the observation `code`** (`types.ObservationRequiresOnsetAndInterpretation`):
  `ancestry` and `consanguinity` are exempt; every other code requires both fields. Keying on the
  value-set-validated `code` (rather than the free-text `system`) makes the rule robust.
- **Requiredness enforced in the worker** (not the binding). The Gin binding cannot express
  "required unless code ∈ {ancestry, consanguinity}", so `onset_code` / `interpretation_code` are no
  longer `required` at binding; the worker applies the rule during validation. This matches how the
  batch already validates codes (against value sets) worker-side. `interpretation_code` still keeps
  its `omitempty,oneof=positive negative` format check at the binding.
- **Ancestry & consanguinity are Radiant-owned value sets.** Both use `coding_system = "radiant"` and
  their observation `value` is validated against their value set. `consanguinity` already existed;
  `ancestry` gets a reference table created **empty** by migration `000015` — its `qc-ethnicity` codes
  are **instance-provided** (seeded per deployment, not shipped in this generic repo). An empty ancestry
  set therefore **rejects** ancestry values (fail-loud); only instance-seeded codes are accepted.
  External-vocabulary observations (phenotype→HPO, condition→MONDO) keep format-only value validation.
- **Worker also validates `interpretation_code`** (previously unchecked) via a new
  `observation_interpretation` value set.
- **Persistence**: `ObsCategorical.OnsetCode` / `InterpretationCode` are now `*string`, persisted as
  `NULL` (via `utils.NilIfEmpty`) instead of `""` — the FK to `onset` / `obs_interpretation` rejects `""`.
- **DB migration** `000015_create_ancestry_value_set` creates the `ancestry` reference table **empty**
  (no seed data shipped — the `qc-ethnicity` codes are instance-seeded). (The onset/interpretation
  columns were already nullable — no migration needed for those.)
- OpenAPI spec + TS/Python clients reflect `onset_code` as non-required (regenerated with
  openapi-generator 7.12.0).

## ☑️ TO-DOs

- [ ] None

## 🧪 Tests

- Worker validation: a required code (e.g. `phenotype`) missing onset/interpretation → errors;
  `ancestry` and `consanguinity` without them → OK; an exempt code with an invalid onset still rejected.
- Value validation: `ancestry` / `consanguinity` with an out-of-set value → error; valid values → OK.
- `types.ObservationRequiresOnsetAndInterpretation`: exempt vs required codes, incl. the empty-code default.
- Handler binding: an invalid `interpretation_code` value → 400; an observation without onset → 202
  (the binding no longer gates onset).
- Integration (Postgres+MinIO): an `ancestry` observation (against a seeded ancestry code) persists with
  `onset_code`/`interpretation_code` = `NULL` (FK satisfied).
- Unit test for the `NilIfEmpty` helper.
- All touched tests use `testutils.RunTest`.

## 🔗 Related Issues

[CLIN-6022](https://ferlab-crsj.atlassian.net/browse/CLIN-6022)

## 👀 Notes for Reviewers

- **Requiredness moved to the worker.** A missing `onset_code` on a required observation is no longer
  a synchronous `400` at ingestion — it surfaces as a batch validation error (status `ERROR` on the
  batch report). This is consistent with the batch's existing worker-side code validation.
- **`required_if` is not expressible in OpenAPI (swaggo)** — the generated spec only marks `onset_code`
  as non-required; the "required unless ancestry/consanguinity" rule lives in the worker.
- Persisting `NULL` (not `""`) is load-bearing: the FK to `onset(code)` / `obs_interpretation(code)`
  would reject an empty string.
- **Ancestry seed intentionally not shipped.** The `qc-ethnicity` codes are QLIN/CQGC-specific, so the
  migration creates the table empty and the codes are seeded **per-instance** (in clin-localstack for
  local dev). Ancestry validation stays enforced against whatever the instance seeds; an unseeded set
  rejects.
- **English-only labels.** Radiant value sets are `code` + `name_en` (no `name_fr`); localization is
  frontend-side. Ancestry codes/labels are instance-seeded (English `display`s from `qc-ethnicity`), not
  shipped here. Adding French labels would be a cross-cutting change (a `name_fr` column across all value
  sets) — out of scope.


backend/internal/types/case_batch.go
```diff
@@ -25,6 +43,8 @@ type CaseBatch struct {
 	Tasks                      []*CaseTaskBatch                 `json:"tasks" toml:"tasks" binding:"required,dive,required"`
 }
 
+// Case Batch
```
celinepelletier: Useless comment

celinepelletier (approved):

backend/scripts/init-sql/migrations/000015_seed_ancestry_value_set.up.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE public.ancestry (
+    code text NOT NULL,
+    name_en text NOT NULL
+);
+
+ALTER TABLE ONLY public.ancestry
+    ADD CONSTRAINT ancestry_pkey PRIMARY KEY (code);
+
+INSERT INTO public.ancestry VALUES ('CA-FR', 'French Canadian');
```
celinepelletier: Prefer 1 single insert : 

```
INSERT INTO public.ancestry VALUES
 ('CA-FR', 'French Canadian'),
('EU', 'European Caucasian'),
....
```

backend/internal/types/ancestry.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+type Ancestry struct {
+	ValueSet
+}
+
+var AncestryTable = Table{
+	Name:           "ancestry",
+	FederationName: "radiant_jdbc.public.ancestry",
```
celinepelletier, line 9: You need to add this in ViewTables in starrocks_tenant repository so it will exists in tenant databases

backend/scripts/init-sql/qlin/ancestry_value_set.sql
```diff
@@ -0,0 +1,6 @@
+-- QLIN-specific ancestry codes, complementing the generalized set from migration 000015.
+-- Applied per-instance (not part of radiant-portal's own seed flow).
+INSERT INTO public.ancestry (code, name_en) VALUES
```
celinepelletier: I would name those files with the table name

backend/scripts/init-sql/qlin/project.sql
```diff
@@ -0,0 +1,7 @@
+-- QLIN-specific clinical projects, complementing the generic RADIANT seed (N1/N2/QLIN).
+-- tenant_code has no default (migration 000013), so it must be set explicitly.
+-- Applied per-instance (not part of radiant-portal's own seed flow).
+INSERT INTO project (code, name, tenant_code) VALUES
```
celinepelletier, line 1: Same

celinepelletier (commented):
Also if you could add organizations and analysis_catalog we added for QLIN in those scripts please

